Issue #12, opened 2022-09-15
`userinfo` method does not handle cases where scope is just `openid`


kwajiehao:
### Describe the bug
`sgid-client` does not handle cases where the client's scope is just `openid`. 

When a client's scope is just `openid`, the auth request is not associated with any data or enc key. Therefore, the response from the sgID user info endpoint will contain undefined `data` and `key` attributes. Because the `sgid-client` does not perform null or `isEmpty` checks on the `key` and `data` attributes, it attempts to import an empty key, resulting in an error.

The error message is also misleading, because `sgid-client` obscures the original error thrown by the `jose` function used to decrypt the private key.

### To Reproduce
1. Create and register a new sgID client with just the `openid` scope
2. Integrate with sgID using the SDK, following the instructions at https://docs.id.gov.sg/onboarding-guide/integrating-using-the-sdk 
3. When retrieving data from the user info endpoint, the library throws an error message `Unable to decrypt or import payload key`

### Expected behavior
`sgid-client` skips the decryption step if the `key` attribute does not exist and returns just the sub.

pregnantboy:
Closed by #14

PR #14, opened 2022-09-29
fix: check that encKey and data are defined before attempting decrypt

kwajiehao:
# Problem
Currently, the `userinfo` method expects to receive a response from the sgid userinfo endpoint that has the type:
```
{
  sub: string
  key: string
  data: Record<string, string>
}
```

However, if you check the [userinfo controller](https://github.com/datagovsg/sgid-server/blob/develop/src/controllers/transaction/userinfo.controller.ts) in the sgid-server repo, this shape is not correct: in particular, the key and data attributes are possibly of type `undefined`. This happens when the client scope is just `openid` (i.e. it doesn't request any data fields). In such cases, the decrypt payload step fails because it attempts to decrypt the payload using a key that is undefined.

This PR addresses issue #12

## Solution
This PR checks that the `encryptedPayloadKey` and `data` are not undefined before attempting to decrypt the payload.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -142,14 +142,17 @@ export class SgidClient {
       key: encryptedPayloadKey,
       data,
     } = await this.sgID.userinfo<{
-      sub: string
-      key: string
-      data: Record<string, string>
+      sub: string | undefined
+      key: string | undefined
+      data: Record<string, string> | undefined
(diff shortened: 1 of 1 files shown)


pregnantboy (approved):
lgtm!

PR #24, opened 2023-03-30
test: set up test suite with Jest and MSW

mantariksh:
## Problem

Tests are good but we don't have them. Also, as we develop new SDKs, it will be useful to have consistent test cases as a reference point for implementation.

## Solution

I considered two possible approaches:
1. Mock `openid-client` and check that we are interacting with it correctly
2. Mock network requests and ensure that the SDK returns the correct outputs

I went with approach 2 because 1 doesn't test for the possibility that `openid-client` introduces a bug or change in behaviour.

I chose `msw` to mock network requests instead of digging into the internals of `openid-client` and mocking whatever HTTP library they use (the native NodeJS HTTP API), as this allows us to stay agnostic to `openid-client`'s internal implementation and just mock API responses.

Also, tests are now part of CI pipeline. See "Test" GitHub action.

**Build changes**
To include the test directory in the TS project but exclude it from the build pipeline, I set `tsconfig.json` to include tests, but created a new `tsconfig.build.json` which excludes tests and used it in `npm run build`.

**New dev dependencies**:

- `jest`, `ts-jest`, `msw`: for test framework in TypeScript
- `jsonwebtoken`: for generating testing data (sign ID token)
- `eslint-plugin-jest`: to write test nice nice


test/SgidClient.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+} from './mocks/handlers'
+import { server } from './mocks/server'
+
+const DEFAULT_SCOPE = 'myinfo.nric_number openid'
+const DEFAULT_RESPONSE_TYPE = 'code'
+
+describe('SgidClient', () => {
+  let client: SgidClient
+
+  beforeEach(() => {
```
PrawiraGenestonlia: I'd use beforeAll instead of beforeEach since the client remains static throughout entire spec and the mock values are constants.

```suggestion
  beforeAll(() => {
```

test/SgidClient.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+    it('should initialise correctly when a PKCS1 private key is provided', () => {
+      expect(client).toBeDefined()
+    })
+
+    it('should initialise correctly when a PKCS8 private key is provided', () => {
+      const pkcs8Key = readFileSync(
+        `${__dirname}/mocks/mockPrivateKeyPkcs8.pem`,
+      ).toString()
+
+      const client = new SgidClient({
```
PrawiraGenestonlia: The naming convention is confusing for scoped client and global client. May not be maintainable in long run.

```suggestion
      const pkcs8Client = new SgidClient({
        clientId: MOCK_CLIENT_ID,
        clientSecret: MOCK_CLIENT_SECRET,
        privateKey: pkcs8Key,
        redirectUri: MOCK_REDIRECT_URI,
        hostname: MOCK_HOSTNAME,
        apiVersion: MOCK_API_VERSION,
      })

      expect(pkcs8Client).toBeDefined()
```

test/SgidClient.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+        clientSecret: MOCK_CLIENT_SECRET,
+        privateKey: pkcs8Key,
+        redirectUri: MOCK_REDIRECT_URI,
+        hostname: MOCK_HOSTNAME,
+        apiVersion: MOCK_API_VERSION,
+      })
+
+      expect(client).toBeDefined()
+    })
+  })
```
PrawiraGenestonlia, line 66: Can consider add another test case for constructor error? i.e. negative test case for when PKCS1/PKCS8 key is not provided. This allows us to see if constructor throws proper error message especially if this is a library that will be used by public.

test/SgidClient.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+      )
+      // Important check is here
+      expect(actual.searchParams.get('redirect_uri')).toBe(mockRedirectUri)
+      expect(actual.searchParams.get('nonce')).toBe(nonce)
+      expect(actual.searchParams.get('state')).toBe(mockState)
+      // Client ID, scope, response_type, redirect_uri, nonce, state
+      expect(actualNumSearchParams).toBe(6)
+    })
+
+    it('should throw when no redirectUri is provided', () => {
```
PrawiraGenestonlia, line 233: current negative test case only covers no redirectUri is provided. Can consider state and scope is not provided respectively. Not super critical but for client facing library, we want to standardise all the error message.

My suggestion is to have all error messages as a constant file. This will help us in documentating this library when being published.

test/SgidClient.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+      server.use(userInfoHandlerNoData)
+
+      const { sub, data } = await client.userinfo(MOCK_ACCESS_TOKEN)
+
+      expect(sub).toBe(MOCK_SUB)
+      expect(data).toEqual({})
+    })
+
+    it('should throw when private key is invalid', async () => {
+      const client = new SgidClient({
```
PrawiraGenestonlia: same as above comment. rename client.

```suggestion
      const invalidClient = new SgidClient({
```

test/SgidClient.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+    it('should throw when sub is empty', async () => {
+      server.use(tokenHandlerNoSub)
+
+      await expect(client.callback(MOCK_AUTH_CODE)).rejects.toThrow(
+        'Missing sub claim or access token',
+      )
+    })
+  })
+
+  describe('userinfo', () => {
```
PrawiraGenestonlia, line 275: great negative test cases for userinfo! :D

test/SgidClient.spec.ts
mantariksh: hmm okay so the constructor doesn't actually throw when the private key isn't provided. currently if the private key is not provided or somehow malformed, the library will only throw when eventually you try to decrypt a payload by calling `userinfo`.

if we want the library to throw, we can tackle that in a separate PR and add a test for it then. currently I'm focusing on testing existing functionality.

test/SgidClient.spec.ts
mantariksh: so currently the library also doesn't throw if the state and scope are not provided. the server may throw, but imo testing that behaviour should be out of scope for testing this client library

test/SgidClient.spec.ts
mantariksh: [8fa730f](https://github.com/opengovsg/sgid-client/pull/24/commits/8fa730fbe476cae80ad3c699c9c941b615a410b9)

test/SgidClient.spec.ts
mantariksh: [a8f7374](https://github.com/opengovsg/sgid-client/pull/24/commits/a8f737474b342dca211714bb454fa727f299ebf4)

test/SgidClient.spec.ts
mantariksh: [a8f7374](https://github.com/opengovsg/sgid-client/pull/24/commits/a8f737474b342dca211714bb454fa727f299ebf4)

mantariksh:
@PrawiraGenestonlia thanks for the thoughtful review! left some of the comments open for discussion

PrawiraGenestonlia (approved):

PR #27, opened 2023-04-12
feat: refine docs and error messages

mantariksh:
Before publishing multiple SDKs, we should set up this one as an example. This PR does the following
1. Refines and adds detail to documentation for each function
2. Extracts all error messages into a separate file, so it is easy for other SDKs to reference and ensure the error messages are consistent
3. Refines error messages to be clearer and (wherever possible) more actionable


LoneRifle (approved):
lgtm

kwajiehao (approved):
lgtm

PR #30, opened 2023-04-17
feat: PKCE `codeVerifier`, `codeChallenge`, and `authorizationUrl`

raynerljm:
## Problem

_What problem are you trying to solve? What issue does this close?_

Closes [insert issue #]

## Solution

_How did you solve the problem?_

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Concerns

3. Should we refactor `authorizationUrl` into its own file? As the new private helper functions `authorizationUrlV1` and `authorizationUrlV2` have been created, the `index.ts` file length has been getting longer and thus less readable (Same for the tests)

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

_What tests should be run to confirm functionality?_

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, etc._

**New environment variables**:

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

- `dependency` : dependency details

**New dev dependencies**:

- `dependency` : dependency details

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,17 +84,17 @@ export class SgidClient {
   ): { url: string; nonce?: string } {
     switch (this.apiVersion) {
       case 1:
-        return this.authorizationUrlV1(state, scope, nonce, redirectUri)
+        return this.authorizationUrlV1({ state, scope, nonce, redirectUri })
 
       case 2:
-        return this.authorizationUrlV2(
+        return this.authorizationUrlV2({
           state,
           scope,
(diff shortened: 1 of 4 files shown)


raynerljm:
Closed as this PR was getting big so I split it up into #32 and #33

PR #31, opened 2023-04-17
feat: PKCE code verifier

raynerljm:
## Problem

Adds the `codeVerifier` function to generate a code verifier for the [PKCE flow](https://github.com/datagovsg/sgid-server/issues/244).

## Concerns 

## Tests

Tests were added in `SgidClient.spec.ts`

Test strategy outline can be found [here](https://www.notion.so/opengov/PKCE-sgid-client-Test-Strategy-a5ad8816c7ca4bbaaa4e5b5f6e199526?pvs=4)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -194,6 +194,26 @@ export class SgidClient {
     }
     return result
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


raynerljm:
Merged this PR with #32 as @PrawiraGenestonlia mentioned that the PRs are a bit too granular

PR #32, opened 2023-04-17
feat: PKCE `codeVerifier` and `codeChallenge` (Option 1.1)

raynerljm:
## Problem

Supporting the [[Proof Key for Code Exchange (PKCE)](https://www.rfc-editor.org/rfc/rfc7636)](https://www.rfc-editor.org/rfc/rfc7636) extension to OAuth 2.0 is a current best practice for preventing authorisation code interception attacks in native apps. As per [[RFC 8252 (OAuth 2.0 for Native Apps)](https://www.rfc-editor.org/rfc/rfc8252)]:

> 

## Solution

This is PR **1** of 3 for [Option 1](https://www.notion.so/opengov/PKCE-sgid-client-breaking-changes-6e2854240f75496f89335d69f709acff?pvs=4)

PR 2 of 3 (#33)
PR 3 of 3 (#34)

**Features**:

Adds the `codeVerifier` and `codeChallenge` function to generate a code challenge for the [PKCE flow](https://github.com/datagovsg/sgid-server/issues/244).

`codeChallenge` is a wrapper over `openid-client` [`generators.codeChallenge`](https://github.com/panva/node-openid-client/blob/main/docs/README.md#generatorscodechallengeverifier)

## Concerns 

## Tests

Tests were added in `SgidClient.spec.ts`

Test strategy outline can be found [here](https://www.notion.so/opengov/PKCE-sgid-client-Test-Strategy-a5ad8816c7ca4bbaaa4e5b5f6e199526?pvs=4)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -196,6 +196,21 @@ export class SgidClient {
   }
 
   /**
+   * Generates a PKCE challenge pair where `codeChallenge` is the generated S256 challenge from `codeVerifier`
+   * @param length The length of the code verifier
+   * @returns The generated challenge pair
+   */
+  generatePkcePair(length = 43): {
+    codeVerifier: string
+    codeChallenge: string
+  } {
+    const codeVerifier = this.generateCodeVerifier(length)
+    const codeChallenge = this.generateCodeChallenge(codeVerifier)
+
+    return { codeVerifier, codeChallenge }
+  }
+
+  /**
    * Generates the code verifier (random bytes encoded in url safe base 64) to be used in the OAuth 2.0 PKCE flow
    * @param length The length of the code verifier to generate (Defaults to 43 if not provided)
    * @returns The generated code verifier
--- a/src/index.ts
+++ b/src/index.ts
@@ -200,7 +200,7 @@ export class SgidClient {
    * @param length The length of the code verifier to generate (Defaults to 43 if not provided)
    * @returns The generated code verifier
    */
-  codeVerifier(length = 43): string {
+  generateCodeVerifier(length = 43): string {
     if (length < 43 || length > 128) {
       // eslint-disable-next-line typesafe/no-throw-sync-func
       throw new Error(
@@ -220,7 +220,7 @@ export class SgidClient {
    * @param codeVerifier The code verifier to calculate the S256 code challenge for
    * @returns The calculated code challenge
    */
-  codeChallenge(codeVerifier: string): string {
+  generateCodeChallenge(codeVerifier: string): string {
     return generators.codeChallenge(codeVerifier)
   }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -194,6 +194,26 @@ export class SgidClient {
     }
     return result
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -194,6 +194,35 @@ export class SgidClient {
...
+  }
+
+  /**
+   * Calculates the S256 PKCE code challenge for a provided code verifier
+   * @param codeVerifier The code verifier to calculate the S256 code challenge for
+   * @returns The calculated code challenge
+   */
+  codeChallenge(codeVerifier: string): string {
+    return generators.codeChallenge(codeVerifier)
+  }
```
PrawiraGenestonlia, line 240: I'd personally merge these two functions into one as they have to be called in pair.  
My thoughts are 
generatePkcePair(length?:number) => 
```
{
    code_verifier: 'u1ta-MQ0e7TcpHjgz33M2DcBnOQu~aMGxuiZt0QMD1C',
    code_challenge: 'CUZX5qE8Wvye6kS_SasIsa8MMxacJftmWdsIA_iKp3I'
}
```

Thoughts? @kwajiehao @mantariksh @kschiew

PrawiraGenestonlia: This discussion will determine the implementation for all other SDKs too.

Some materials that I am referencing: https://pypi.org/project/pkce/
https://www.npmjs.com/package/pkce-challenge

raynerljm: I think this is a good idea - do we still want to expose `codeChallenge` and `codeVerifier` or only expose the merged function though?

kschiew: From the users' perspective, I think it's a good idea to have a `generatePkcePair` function that generates both the challenge and verifier.

Personally, I agree with what @mantariksh said in the implementation sketch, which is to provide good defaults but also flexibility. We can make it clear in the documentation that using our `generatePkcePair` is strongly recommended, but we should also expose the two functions in the event they have some weird use case(s).

Referencing the two implementations mentioned by @PrawiraGenestonlia, this seems to be what they did as well.

src/index.ts
```diff
@@ -194,6 +194,35 @@ export class SgidClient {
     }
     return result
   }
+
+  /**
+   * Generates the code verifier (random bytes encoded in url safe base 64) to be used in the OAuth 2.0 PKCE flow
+   * @param length The length of the code verifier to generate (Defaults to 43 if not provided)
+   * @returns The generated code verifier
+   */
+  codeVerifier(length = 43): string {
```
kschiew: nit: Shall we name the functions with a verb prefix, eg. `get...()` or `generate...()` for clearer intentions? I understand the underlying library named them without verb prefixes, but I think the name of the class (`generator`) has conveyed the intention of the function clearly already, which might not be case in our situation.

PrawiraGenestonlia: agree with @kschiew. Good to start function name with verb

raynerljm: okay, thanks for the comments guys! will make the changes

src/index.ts
raynerljm: yup got it, that makes a lot of sense - thanks!

raynerljm:
Closed as [Option 3](https://github.com/opengovsg/sgid-client/pull/35) was selected

PR #35, opened 2023-04-19
feat: PKCE

### LINKED ISSUE #26: Implement code verifier for PKCE
See issue https://github.com/datagovsg/sgid-server/issues/244

raynerljm:
## Problem

Supporting the [[Proof Key for Code Exchange (PKCE)](https://www.rfc-editor.org/rfc/rfc7636)](https://www.rfc-editor.org/rfc/rfc7636) extension to OAuth 2.0 is a current best practice for preventing authorisation code interception attacks in native apps. As per [[RFC 8252 (OAuth 2.0 for Native Apps)](https://www.rfc-editor.org/rfc/rfc8252)](https://www.rfc-editor.org/rfc/rfc8252):

> 

### High-level principles

As this is a relatively large PR which involves breaking changes, I would like to highlight a few high-level principles that were kept in mind during development and which should be achieved through the code changes.

1. The SDK should be dummy

This is in line with what we have been doing so far where we would like to keep the SDK dummy. This is so that if we would like to make any changes (e.g. with validation), we are able to modify our servers directly instead of updating our SDK (as not all RPs will update their SDKs).

This is achieved by not validating the ABNF of the codeChallenge and codeVerifier received in `authorizationUrl` and `callback` respectively. (Especially because validation bugs are easier to commit and harder to rectify)

2. We should be mindful of the exports we are providing

As we are maintaining a public SDK for people to use, we need to be very aware of what exactly we are exposing for developers to import. If we export function `a`, we have to continue maintaining `a` as an export in future versions as developers might be relying on them.

This is achieved by refactoring the `SgidClient` class into it's own file and keeping `index.ts` solely as a tunnel for exports. Additionally, we are now exporting the SDK types so developers with TypeScript can make use of them.

3. We should be mindful of the SDK function signature

With this PR, the API of the SDK is changing (i.e.  the function signature is different). Specifically, the functions now take in an <Object> instead of sequential parameters.

This is because as the list of parameters gets longer this (i) makes it harder to remember which input corresponds to which parameter and (ii) makes it very confusing when there are optional parameters - users would have to enter `undefined` for the optional parameters they do not want to fill in just to reach the parameter they have to fill in.

**Features**:

PKCE related changes
- Added the `generatePkcePair`, `generateCodeVerifier` and `generateCodeChallenge` functions based on `openid-client` `generator` functions in the `generators.ts` file
- Changed the `authorizationUrl` function params to take in an object
    - Added the `codeChallenge` ~~and `codeChallengeMethod`~~ parameter
- Changed the `callback` function params to take in an object
    - Added the `codeVerifier` parameter
- Added relevant tests for new functions
    - Added tests for `generatePkcePair`, `generateCodeVerifier` and `generateCodeChallenge` functions
    - Updated tests for `authorizationUrl` and `callback`
    
SDK improvement related changes
- Changed the `userinfo` function params to take in an object
- Changed default scope from `openid myinfo.nric_number` to `openid myinfo.name` 
- Refactored main `SgidClient` class into its own file `sgidClient.ts` and extracted types into `types.ts`
- Exported `sgidClient.ts`, `types.ts`, `generators.ts` and `util.ts` through `index.ts` (to be explicit about what we want exported)
- Removed `apiVersion` as a parameter for the constructor and use a hardcoded `API_VERSION` constant
- Updated the issuer to have the `v2` appended (in the constructor)
- ~~Turn off~~ Remove `eslint-no-throw-sync-func` ESLint rule 

README related changes
- Added description + example usage for generatePkcePair
- Added function signature (parameters + return type) indicating the type, whether it's optional, the default value, and a brief explanation
- Removed the code snippet inline comments to the function signature

Non-implemented changes
~~- Changing the constructor param from `redirectUri` to `redirectUris`~~

## Tests
- Added test suites for `generatePkcePair`, `generateCodeVerifier`, and `generateCodeChallenge`
- Updated test cases for `authorizationUrl` and `callback` functions
- Tested E2E locally following [this flow](https://www.notion.so/opengov/Testing-flow-from-Test-App-thru-Node-js-SDK-to-Local-server-8e9c301e7e36403d95fcfc02f9dedf8b?pvs=4)

**Verified that all tests are passing**
image


src/index.ts
```diff
@@ -67,20 +90,39 @@ export class SgidClient {
    * @param scopes Array or space-separated scopes, must include openid
    * @param nonce Specify null if no nonce
    * @param redirectUri The redirect URI used in the authorization request, defaults to the one registered with the client
+   * @param codeChallenge The code challenge from the code verifier used for PKCE enhancement
+   * @param codeChallengeMethod The code challenge method used to generate the code challenge from the code verifier, must be `S256`
```
kschiew: Are we validating the value of `codeChallengeMethod` is indeed 'S256' anywhere in the function? 

Although validation is also carried out in the `sgid-server`, but imo it's also good practice to just validate it in our SDK code, especially we are doing so for other variables such as `codeChallenge`.

src/util.ts
```diff
@@ -7,3 +7,9 @@ export function convertPkcs1ToPkcs8(pkcs1: string): string {
   const key = new NodeRSA(pkcs1, 'pkcs1')
   return key.exportKey('pkcs8')
 }
+
+/**
+ * Regex pattern that the code verifier and code challenge in the PKCE flow should match according to the PKCE RFC
+ * https://www.rfc-editor.org/rfc/rfc7636
+ */
+export const codeVerifierAndChallengePattern = /^[A-Za-z\d\-._~]{43,128}$/
```
kschiew: Suggestion: Should we put this under `constants.ts` instead? Especially when we are not using this regex pattern directly in this file anyways.

src/constants.ts
```diff
@@ -0,0 +1,9 @@
+import { ClientAuthMethod, ResponseType } from 'openid-client'
+
+export const SGID_SIGNING_ALG = 'RS256'
```
kschiew, line 3: Suggestion: Should we have block comments to group our constants? What I had in mind was something similar to how we group constants in `sgid-server` - Could be just a phrase, eg. 'PKCE', 'API versioning' etc.

src/index.ts
```diff
@@ -1,51 +1,69 @@
...
+      token_endpoint: `${hostname}/v${this.apiVersion}/oauth/token`,
+      userinfo_endpoint: `${hostname}/v${this.apiVersion}/oauth/userinfo`,
       jwks_uri: `${new URL(hostname).origin}/.well-known/jwks.json`,
     })
 
     this.sgID = new Client({
       client_id: clientId,
       client_secret: clientSecret,
-      redirect_uris: redirectUri ? [redirectUri] : undefined,
+      redirect_uris: redirectUris,
```
kschiew: Just wanna say good job for fixing this! Courtesy of @PrawiraGenestonlia also of course

src/index.ts
```diff
@@ -104,17 +146,33 @@ export class SgidClient {
...
+  }: CallbackParams): Promise<{ sub: string; accessToken: string }> {
+    if (this.apiVersion !== 2) {
+      // eslint-disable-next-line typesafe/no-throw-sync-func
+      throw new Error(
+        "Code verifier must be provided in 'callback' when using apiVersion 2",
+      )
+    }
+
+    if (codeVerifier === undefined) {
+      // eslint-disable-next-line typesafe/no-throw-sync-func
```
kschiew: Question: I notice we disable this specific lint rule every time we throw an Error in the codebase - should we just refactor and disable this rule in config instead?

src/index.ts
```diff
@@ -194,6 +252,50 @@ export class SgidClient {
...
+
+  /**
+   * Calculates the S256 PKCE code challenge for a provided code verifier
+   * @param codeVerifier The code verifier to calculate the S256 code challenge for
+   * @returns The calculated code challenge
+   */
+  static generateCodeChallenge(codeVerifier: string): string {
+    return generators.codeChallenge(codeVerifier)
+  }
 }
```
kschiew: Suggestion: Should we put these functions in a separate `utils` file instead?

kschiew (changes requested):
Overall LGTM! Just a few suggestions on organising the code

src/index.ts
raynerljm: yeah so i'm unsure about this because i omitted the check in line with trying to keep the SDK 'dumb' so that we can make changes to the server at our own cadence.

would this be something we want to check?

src/util.ts
raynerljm: yup got it

src/constants.ts
raynerljm: yup okay that makes sense

src/index.ts
raynerljm: yeah i think i'll do that since this has been occurring multiple times (unless someone has objections)

Update: talked to Alwyn about it - seems like it MIGHT be something we want to keep - will discuss with @kwajiehao in our one-on-one

src/index.ts
raynerljm: yeah i'm thinking about refactoring these into a `generators.ts` file

src/util.ts
raynerljm: Refactored in this [PR](https://github.com/opengovsg/sgid-client/pull/39) in order to prevent this PR from growing too large

src/constants.ts
raynerljm: Refactored in this [PR](https://github.com/opengovsg/sgid-client/pull/39) in order to prevent this PR from growing too large

src/index.ts
raynerljm: Refactored in this [PR](https://github.com/opengovsg/sgid-client/pull/39) in order to prevent this PR from growing too large

src/index.ts
raynerljm: after some discussion we've decided that this ESLint rule, while it does have its use, is quite opinionated and not necessarily considered a standard practice (many Python and Node libraries throw errors in synchronous functions) and as such we will be removing it

additionally, as our other codebases are not adhering to this specific ESLint rule, we've decided not to adhere to it as part of an effort to standardize our coding practices

src/index.ts
raynerljm: after discussion with Jie Hao, we've decided NOT to include the `codeChallengeMethod` as a parameter. as there are currently only 2 values `plain` and `S256`, and the PKCE RFC clearly highlights that `plain` is insecure, there is no reason to let developer shoot themselves in the foot
(i.e. if we ONLY let them choose `S256`, whats the point of the param? OR if we let them choose between `plain` and `S256` but if they choose `plain` then we throw an error, whats the point of letting them choose?)

from a UX perspective, this reduces the mental load as users do not have to think about what code challenge method to employ

for users that are familiar with the PKCE code challenge methods, we will document on the developer docs that we are ONLY using `S256` as default due to security issues with using `plain`

src/index.ts
```diff
@@ -59,6 +76,11 @@ export class SgidClient {
     } else {
       this.privateKey = privateKey
     }
+
+    /**
+     * To rectify instances where an extra '\' escape character is added to the newline
+     */
+    this.privateKey = this.privateKey.replace(/\\n/gm, '\n')
```
kwajiehao: nice!

src/index.ts
```diff
@@ -67,20 +89,35 @@ export class SgidClient {
...
-    redirectUri: string = this.getFirstRedirectUri(),
-  ): { url: string; nonce?: string } {
+  authorizationUrl({
+    state,
+    scope = DEFAULT_SCOPE,
+    nonce = generators.nonce(),
+    redirectUri = this.getFirstRedirectUri(),
+    codeChallenge,
+  }: AuthorizationUrlParams): AuthorizationUrlReturn {
+    if (this.apiVersion !== 2) {
```
kwajiehao: Why do we need this line of validation if we're setting the api version ourselves?

src/index.ts
```diff
@@ -104,17 +140,31 @@ export class SgidClient {
...
-    code: string,
-    nonce: string | null = null,
+  async callback({
+    code,
+    nonce = null,
     redirectUri = this.getFirstRedirectUri(),
-  ): Promise<{ sub: string; accessToken: string }> {
+    codeVerifier,
+  }: CallbackParams): Promise<{ sub: string; accessToken: string }> {
+    if (this.apiVersion !== 2) {
```
kwajiehao: See above comment

src/index.ts
```diff
@@ -104,17 +140,31 @@ export class SgidClient {
...
+      throw new Error(
+        "Code verifier must be provided in 'callback' when using apiVersion 2",
+      )
+    }
+
+    if (codeVerifier === undefined) {
+      throw new Error(
+        "Code verifier must be provided in 'callback' when using apiVersion 2",
```
kwajiehao: Since we're not giving them a choice to choose the api version, I'm not sure if we need to mention it here. what do you think?

src/index.ts
```diff
@@ -194,6 +244,49 @@ export class SgidClient {
...
+   * Calculates the S256 PKCE code challenge for a provided code verifier
+   * @param codeVerifier The code verifier to calculate the S256 code challenge for
+   * @returns The calculated code challenge
+   */
+  static generateCodeChallenge(codeVerifier: string): string {
+    return generators.codeChallenge(codeVerifier)
+  }
 }
 
 export default SgidClient
```
kwajiehao, line 3: nit: did we have a discussion on whether the client should be exported by default? if i recall, we also mentioned Panva's `openid-client` which also uses named exports. If so, generators can be its own class like in `openid-client`:

```
import { generators } from 'openid-client';
```

Thoughts?

kwajiehao (commented):
left some comments

src/index.ts
kwajiehao: Ah sorry, just saw that you did it in the refactor PR!

src/index.ts
raynerljm: yeah i encountered this issue while setting up the react-express demo and i asked @PrawiraGenestonlia and he gave me this suggestion! :D

PrawiraGenestonlia: Yeah. We should be handling all these edge cases to ensure smooth devX.

src/index.ts
raynerljm: ah okay previously the `apiVersion` was still a parameter in the constructor with the default changed to `2`.

after removing `apiVersion` to be a constant (`API_VERSION`) that users cannot control, i forgot to remove these checks. will do so now

src/index.ts
raynerljm: actually since the `codeVerifier` should no longer be optional with PKCE, do you think we should make it not optional in the type signature and also remove this check?

raynerljm: same for the check for `codeChallenge === undefined` above in the `authorizationUrl` function (i made this change first so we can see it in the code - i'll revert the commit if we decide to keep the checks'

raynerljm: same for the check for `codeChallenge === undefined` above in the `authorizationUrl` function

.eslintrc
```diff
@@ -60,7 +60,7 @@
         "import/no-duplicates": "error",
         "@typescript-eslint/no-floating-promises": 2,
         "@typescript-eslint/no-unused-vars": 2,
-        "typesafe/no-throw-sync-func": "error"
+        "typesafe/no-throw-sync-func": "off"
```
PrawiraGenestonlia: why?

src/index.ts
```diff
@@ -67,20 +89,25 @@ export class SgidClient {
...
-    state: string,
-    scope: string | string[] = 'myinfo.nric_number openid',
-    nonce: string | null = generators.nonce(),
-    redirectUri: string = this.getFirstRedirectUri(),
-  ): { url: string; nonce?: string } {
+  authorizationUrl({
+    state,
+    scope = DEFAULT_SCOPE,
+    nonce = generators.nonce(),
+    redirectUri = this.getFirstRedirectUri(),
```
PrawiraGenestonlia: can we do a check if redirectUri is in this.redirectUris?

PrawiraGenestonlia (changes requested):
left some comments

src/index.ts
```diff
@@ -1,51 +1,68 @@
...
     // TODO: Discover sgID issuer metadata via .well-known endpoint
     const { Client } = new Issuer({
       issuer: new URL(hostname).origin,
-      authorization_endpoint: `${hostname}/v${apiVersion}/oauth/authorize`,
-      token_endpoint: `${hostname}/v${apiVersion}/oauth/token`,
-      userinfo_endpoint: `${hostname}/v${apiVersion}/oauth/userinfo`,
+      authorization_endpoint: `${hostname}/v${this.apiVersion}/oauth/authorize`,
+      token_endpoint: `${hostname}/v${this.apiVersion}/oauth/token`,
+      userinfo_endpoint: `${hostname}/v${this.apiVersion}/oauth/userinfo`,
       jwks_uri: `${new URL(hostname).origin}/.well-known/jwks.json`,
```
kschiew: Updates due to our OpenID Discovery changes:
1. New Issuer will be appended with `/v2`.
2. That implies for our OAuth and OpenID endpoints, we don't have to manually insert `/v2` anymore.

The server-side changes can be found [here](https://github.com/datagovsg/sgid-server/pull/445).

kschiew (changes requested):
New changes needed due to OpenID Discovery endpoint changes!

src/index.ts
```diff
@@ -1,51 +1,68 @@
...
+
+type CallbackParams = {
+  code: string
+  nonce?: string | null
+  redirectUri?: string
+  codeVerifier: string
+}
 
 export class SgidClient {
+  private apiVersion: number
```
mantariksh: why do we need to make this a class member when it's not used outside the constructor?

src/index.ts
```diff
@@ -1,51 +1,68 @@
...
+      token_endpoint: `${hostname}/v${this.apiVersion}/oauth/token`,
+      userinfo_endpoint: `${hostname}/v${this.apiVersion}/oauth/userinfo`,
       jwks_uri: `${new URL(hostname).origin}/.well-known/jwks.json`,
     })
 
     this.sgID = new Client({
       client_id: clientId,
       client_secret: clientSecret,
-      redirect_uris: redirectUri ? [redirectUri] : undefined,
+      redirect_uris: redirectUris,
```
mantariksh: actually, I'm not so sure what this adds to the library.

even if we take multiple redirect URIs in the constructor, we don't know which one to use later on in the other functions. so even if the user passes multiple redirect URIs in the constructor, they have to pass one of them again later on. so why pass them in the constructor?

if the point is to pass an array of redirect URIs initially, then later on when the user passes in another redirect URI, we validate that this one is part of the original array...it doesn't seem to add much value? we're not really validating anything if we're validating user input against other user input.

imo the point of passing a redirect URI in the constructor is to act as a default for the majority of cases where there's only one redirect URI, so you can pass all the config in one place and not worry about it later on. in which case it makes sense to pass just one.

src/index.ts
```diff
@@ -1,51 +1,68 @@
...
+  SGID_AUTH_METHOD,
+  SGID_SIGNING_ALG,
+  SGID_SUPPORTED_FLOWS,
+} from './constants'
 import { convertPkcs1ToPkcs8 } from './util'
 
-const SGID_SIGNING_ALG = 'RS256'
-const SGID_SUPPORTED_FLOWS: ResponseType[] = ['code']
-const SGID_AUTH_METHOD: ClientAuthMethod = 'client_secret_post'
+type ConstructorParams = {
```
mantariksh: wdyt about exposing all these types? I've frequently found it very helpful when libraries expose types like these, so I can easily write the types for helper functions e.g. to construct the required args to the pass to the library. and there's not much downside that I can see

test/SgidClient.spec.ts
```diff
@@ -56,26 +59,29 @@ describe('SgidClient', () => {
...
-        apiVersion: MOCK_API_VERSION,
       })
 
       expect(pkcs8Client).toBeDefined()
     })
   })
 
-  describe('authorizationUrl', () => {
-    it('should generate authorisation URL correctly when only state is provided', () => {
+  describe('authorizationUrl (v2 with PKCE)', () => {
```
mantariksh: minor point but I'm not sure it's helpful to add all this v2 naming if we're not keeping v1 anyway. it's a little confusing

test/SgidClient.spec.ts
```diff
@@ -314,19 +371,68 @@ describe('SgidClient', () => {
...
       await expect(client.userinfo(MOCK_ACCESS_TOKEN)).rejects.toThrow(
         'Unable to decrypt payload',
       )
     })
   })
+
+  describe('generatePkcePair', () => {
+    it('can generate a PKCE pair', () => {
+      const pkcePair = SgidClient.generatePkcePair()
+      expect(pkcePair).toBeDefined()
```
mantariksh: feel like we can be more specific here - we expect the pair to have a specific format (specific keys, strings which match some pattern) right? if so we should test that the format is correct.

I know the code in `generatePkcePair` relies on `generateCodeVerifier` and `generateCodeChallenge` internally, and the latter 2 are tested below. but we shouldn't make assumptions about the implementation in case it changes in future.

test/mocks/handlers.ts
```diff
@@ -26,35 +27,39 @@ const jwksHandler = rest.get(MOCK_JWKS_ENDPOINT, (_req, res, ctx) => {
...
+    )
+  },
+)
 
 /**
  * Handler to test case where server doesn't return access token
  */
-export const tokenHandlerNoToken = rest.post(
-  MOCK_TOKEN_ENDPOINT,
+export const tokenHandlerNoTokenV2 = rest.post(
```
mantariksh: again, not sure there's value in changing all these to v2 if we're not keeping v1. makes things confusing for future developers

.eslintrc
raynerljm: @kwajiehao and i had an earlier [discussion](https://github.com/opengovsg/sgid-client/pull/35#discussion_r1177290124) about whether we want to enforce this practice of not throwing errors in (synchronous) function

conclusion was

> after some discussion we've decided that this ESLint rule, while it does have its use, is quite opinionated and not necessarily considered a standard practice (many Python and Node libraries throw errors in synchronous functions) and as such we will be removing it

> additionally, as our other codebases are not adhering to this specific ESLint rule, we've decided not to adhere to it as part of an effort to standardize our coding practices

feel free to respond if you would like to reopen and contribute to the discussion

src/index.ts
raynerljm: in the previous implementation (where access to v1 and v2 coexists in the SDK), a check was being done to ensure that the `apiVersion` must be `2` when constructing an authorization url with the `codeChallenge`. 

that has since been removed and thus `apiVersion` is now only being used in the constructor. i'll remove this attribute now as part of the cleanup and reference the constant itself (instead of the constructor params) for the endpoint construction - thanks!

src/index.ts
raynerljm: hmm i do agree with you that it doesn't really 'add' anything per se.

i went to dig into panva's `openid-client` npm library that we're using under the hood because his `Client` constructor also takes in multiple `redirect_uris`. from what i understand [here](https://github.com/panva/node-openid-client/blob/d571ceb2e74cb703467cf717b4582627f2dcc1af/lib/helpers/client.js#L148), it seems like if 1 URI was provided during construction and NO `redirect_uri` was specified while calling `authorizationUrl`, it will default to that 1 URI. otherwise if `redirect_uris.length !== 1`, it will use whatever `redirect_uri` was provided while calling the respective function without doing any form of validation.

i guess logically it makes sense to not change `redirect_uri` to `redirect_uris` but to stay consistent with other popular libraries changing it might make sense

(am discussing this with @PrawiraGenestonlia now)

src/index.ts
raynerljm: yeah that's a great idea! let's do that. i also realized that while migrating Alwyn's demo app for LifeHack form JS to TS that i needed to define the types manually

.eslintrc
PrawiraGenestonlia: yeah.. I mean just remove it instead of turning off if not required.

src/index.ts
raynerljm: okay after discussing with @PrawiraGenestonlia we've decided to leave the current implementation as is because there is no upside to allowing RPs to register multiple `redirectUris` and might instead cause more confusion.

i will update the comment on the `authorizationUrl` function to describe how the `redirectUri` passed in will override the one passed in during the constructor

test/mocks/handlers.ts
raynerljm: mm that makes total sense! will do 👍

test/SgidClient.spec.ts
raynerljm: yup that makes sense! will make the change 👍

test/SgidClient.spec.ts
raynerljm: yeah i agree that the tests shouldn't make assumptions about the implementation - totally valid point that slipped my mind. thanks so much! i'll make the changes

src/index.ts
raynerljm: now that we've reverted from `redirectUris` (a string array) to just `redirectUri` (a single string input), this is no longer necessary

src/index.ts
raynerljm: okay thank you! as we discussed in DMs:
- since we are not currently using the well known endpoints but instead manually constructing our endpoints, change 2 does not apply
- change 1 has still been made and `v2` is appended to the `issuer` field

.eslintrc
raynerljm: ahh that makes sense - done! thank you :D

src/types.ts
```diff
@@ -0,0 +1,28 @@
+export type AuthorizationUrlParams = {
+  state?: string
+  scope?: string | string[]
+  nonce?: string | null
```
kschiew, line 4: Not super familiar with this part of TypeScript, but is there a reason we still need `| null` when we already have `nonce?` as an optional field?

src/sgidClient.ts
```diff
@@ -0,0 +1,229 @@
...
+  constructor({
+    clientId,
+    clientSecret,
+    privateKey,
+    redirectUri,
+    hostname = 'https://api.id.gov.sg',
+  }: SgidClientParams) {
+    // TODO: Discover sgID issuer metadata via .well-known endpoint
+    const { Client } = new Issuer({
+      issuer: `${new URL(hostname).origin}/v${API_VERSION}`,
```
kschiew: Note that the original issuer does not have '/v1' appended to it

src/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  redirectUri?: string
+  codeChallenge: string
+}
+
+export type AuthorizationUrlReturn = { url: string; nonce?: string }
+
+export type CallbackParams = {
+  code: string
+  nonce?: string | null
```
kschiew, line 13: Same as comment above

test/mocks/constants.ts
```diff
@@ -7,8 +7,8 @@ import mockClientKeys from './mockClientKeys.json'
 export const MOCK_PRIVATE_KEY = readFileSync(
   `${__dirname}/mockPrivateKey.pem`,
 ).toString()
-export const MOCK_HOSTNAME = 'https://id.sgid.com'
-export const MOCK_API_VERSION = 3
+export const MOCK_API_VERSION = 2
+export const MOCK_HOSTNAME = `https://id.sgid.com/v${MOCK_API_VERSION}`
 export const MOCK_AUTH_ENDPOINT = `${MOCK_HOSTNAME}/v${MOCK_API_VERSION}/oauth/authorize`
 export const MOCK_TOKEN_ENDPOINT = `${MOCK_HOSTNAME}/v${MOCK_API_VERSION}/oauth/token`
 export const MOCK_USERINFO_ENDPOINT = `${MOCK_HOSTNAME}/v${MOCK_API_VERSION}/oauth/userinfo`
```
kschiew, line 15: Seems to me the endpoints will contain two '/v2' substrings, eg. `MOCK_AUTH_ENDPOINT` would be 'https://id.sgid.com/v2/v2/oauth/authorize'

src/types.ts
kschiew: Just did my research and turns out optional params will still throw error if `null` is provided. Leaving this intentional comment here as a learning point for myself!

kschiew (commented):
Sound logic! Left a few comments regarding URL endpoints

src/types.ts
raynerljm: yup haha because sometime there's occasion to leave the parameter empty or intentionally leave it as `null | undefined`. i think in this case it's for users who want to explicitly leave the nonce as null

src/sgidClient.ts
raynerljm: ahh good catch! will fix that

src/types.ts
PrawiraGenestonlia: ? refers to undefined. null !== undefined

test/mocks/constants.ts
raynerljm: ah yes good catch! fixed :D

src/sgidClient.ts
```diff
@@ -43,12 +43,17 @@ export class SgidClient {
     hostname = 'https://api.id.gov.sg',
   }: SgidClientParams) {
     // TODO: Discover sgID issuer metadata via .well-known endpoint
+    const issuer =
+      (API_VERSION as number) === 1
+        ? new URL(hostname).origin
+        : `${new URL(hostname).origin}/v${API_VERSION}`
```
kschiew: super nit: What do you think of extracting `new URL(hostname).origin` out from the ternary operation? ie.

``const issuer = (new URL(hostname).origin) + (API_VERSION as number) === 1 ? '' : `/v${API_VERSION)` ``

kschiew (commented):
LGTM aside from the nit! Will leave it to other senior engs to review before approving

src/sgidClient.ts
raynerljm: hmm i think it's a good choice to refactor - however, i decided to do it in another way to be more explicit that only version 2 onwards includes the suffix

```javascript
    let issuer = new URL(hostname).origin
    if ((API_VERSION as number) !== 1) {
      issuer += `/v${API_VERSION}`
    }
```

feel free to let me know what you think!

src/index.ts
```diff
@@ -1,228 +1,7 @@
...
-    return result
-  }
-}
+import { SgidClient } from './sgidClient'
 
 export default SgidClient
+export * from './sgidClient'
+export * from './types'
+export * from './util'
+export * from './generators'
```
kwajiehao: since we are changing the API of the SDK, can we update:
1. The repo README
2. The PR description on why we decided to do so

src/sgidClient.ts
```diff
@@ -0,0 +1,237 @@
...
+    hostname = 'https://api.id.gov.sg',
+  }: SgidClientParams) {
+    // TODO: Discover sgID issuer metadata via .well-known endpoint
+    /**
+     * Issuer is appended with version number only from v2 onwards
+     */
+    let issuer = new URL(hostname).origin
+    if ((API_VERSION as number) !== 1) {
+      issuer += `/v${API_VERSION}`
+    }
```
kwajiehao: nit: not sure if this check is necessary, since we will be releasing this as v2

src/sgidClient.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    const { Client } = new Issuer({
+      issuer,
+      authorization_endpoint: `${hostname}/v${API_VERSION}/oauth/authorize`,
```
kwajiehao: What's the rationale for using `${hostname}/v${API_VERSION}` instead of just `issuer`

kwajiehao (commented):
Looks good to me in general, left a few comments

Will leave it to @mantariksh to have the final say on PR approval

src/sgidClient.ts
raynerljm: Discussed separately: Conclusion is that, while there are some instances where (hostname + version_number) != (issuer), it doesn't apply to us and issuer should always == hostname + version_number. 

Changed to from hostname + version_number to `issuer`

src/sgidClient.ts
raynerljm: mm fair point, removed

src/index.ts
raynerljm: Updated the PR description and updated the README with the new function signatures (mostly based off what has been done on [GitBook](https://app.gitbook.com/s/pBW92htBuXTrDYoKovQ2/~/changes/1/sdks/node.js))

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
 
 ```typescript
-const { url } = client.authorizationUrl(
-  'state',
-  ['openid', 'myinfo.nric_number'], // or space-concatenated string
-  null, // defaults to randomly generated nonce if unspecified
-  'http://localhost:3000/other_callback', // overrides redirect uri
-)
+const { url } = client.authorizationUrl({
+  state: 'state', // no longer mandatory as PKCE protects against CSRF
```
PrawiraGenestonlia: Instead of no longer mandatory, maybe "Optional"

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
 
 ### Token exchange
 
-`async client.callback(code, [nonce], [redirectUri])`
+`async client.callback(parameters)`
+
+- parameters: `<Object>`
+  - code: `<string>`
+  - nonce: `<string>`
+  - redirectUri: `<string>`
```
PrawiraGenestonlia: Mark those optional ones? across all parameters..

PrawiraGenestonlia (commented):
Left some comments

PrawiraGenestonlia (approved):
lgtm!

Will leave it to @mantariksh to have the final say on PR approval

.eslintrc
```diff
@@ -59,8 +59,7 @@
         "import/newline-after-import": "error",
         "import/no-duplicates": "error",
         "@typescript-eslint/no-floating-promises": 2,
-        "@typescript-eslint/no-unused-vars": 2,
-        "typesafe/no-throw-sync-func": "error"
```
mantariksh, line 63: hahaha poor yuanruo (he wrote this plugin lol)

README.md
```diff
@@ -18,10 +18,12 @@ npm i @opengovsg/sgid-client
 
 ## Usage
```
mantariksh, line 15: For L9-11 - can we shift the changelog info to the bottom of the README? I think installation and quickstart are much more important

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
 })
 ```
 
+### Generate PKCE pair
+
+`generatePkcePair([length])`
+
+```typescript
+const { codeChallenge, codeVerifier } = generatePckePair()
```
mantariksh:
```suggestion
const { codeChallenge, codeVerifier } = generatePkcePair()
```

README.md
```diff
@@ -18,10 +18,12 @@ npm i @opengovsg/sgid-client
...
+For more detailed instructions on how to register your client and integrate with sgID, please refer to our [developer documentation](https://docs.id.gov.sg/).
+
 ### Initialization
 
 ```typescript
-import SgidClient from '@opengovsg/sgid-client'
+import { SgidClient } from '@opengovsg/sgid-client'
 
 const client = new SgidClient({
   clientId: 'CLIENT-ID',
```
mantariksh: nit: for all these, might be helpful to change the samples to stuff like `<Insert your client ID>` or `<Insert your private key as a string>` so it's clear that users have to replace these with their own values

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
 })
 ```
 
+### Generate PKCE pair
```
mantariksh: wondering whether we should be more cognisant of users who may be looking at this with no prior knowledge of OIDC.

e.g. we can minimise the number of unfamiliar terms. this heading can be "Generate code challenge and verifier" so we avoid mentioning PKCE, and stick to just "code challenge" and "code verifier".

e.g. we can handhold users a little more. if I wasn't familiar with OIDC and I didn't know what this was, I would probably just skip it. and even if I didn't skip it, I wouldn't know that I need to store the code verifier. so maybe we can include a line below along the lines of "This is a required step for OIDC ([more info](https://oauth.net/2/pkce/)). You should store the `codeVerifier` in the user's session so you can retrieve it later."

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
+`generatePkcePair([length])`
+
+```typescript
+const { codeChallenge, codeVerifier } = generatePckePair()
+```
+
 ### Get Authorization URL
 
-`client.authorizationUrl(state, scope, [nonce], [redirectUri])`
+`async client.authorizationUrl(parameters)`
```
mantariksh: this is not async right?

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
 ```typescript
-const { url } = client.authorizationUrl(
-  'state',
-  ['openid', 'myinfo.nric_number'], // or space-concatenated string
-  null, // defaults to randomly generated nonce if unspecified
-  'http://localhost:3000/other_callback', // overrides redirect uri
-)
+const { url } = client.authorizationUrl({
+  codeChallenge: 'zaqUHoBV3rnhBF2g0Gkz1qkpEZXHqi2OrPK1DqRi-Lk', // generated from the previous step
+  state: 'state', // optional as PKCE protects against CSRF
```
mantariksh: we can also reduce the number of unfamiliar terms used here. if user doesn't need to worry about CSRF, we don't need to mention it. instead, we can say something like "for you to track application state which will be passed back to you after login"

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
+  scope: ['openid', 'myinfo.name'], // or space-concatenated string
+  nonce: null, // defaults to randomly generated nonce if unspecified
```
mantariksh: do we want to encourage users to use the nonce? p sure most users won't if we put this in the README

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
+  - redirectUri: `<string>` (Optional)
 
 ```typescript
-const { sub, accessToken } = await client.callback(
-  'code', // auth code reuturned from redirect_url
-  null,
-  'http://localhost:3000/other_callback', // optional, unless overridden
-)
+const { sub, accessToken } = await client.callback({
+  code: 'code', // auth code reuturned from redirect_url
```
mantariksh: might also be helpful to say something like "auth code returned in query params at `redirectUri`"

```suggestion
  code: 'code', // auth code returned from redirect_url
```

README.md
```diff
@@ -31,38 +33,64 @@ const client = new SgidClient({
...
+  - nonce: `<string>` (Optional)
+  - redirectUri: `<string>` (Optional)
 
 ```typescript
-const { sub, accessToken } = await client.callback(
-  'code', // auth code reuturned from redirect_url
-  null,
-  'http://localhost:3000/other_callback', // optional, unless overridden
-)
+const { sub, accessToken } = await client.callback({
```
mantariksh, line 95: might be helpful to write a comment here mentioning what the `sub` is

mantariksh (commented):
starting with comments on the README first!

README.md
PrawiraGenestonlia: what a beast! everyone misses that...

src/generators.ts
```diff
@@ -0,0 +1,48 @@
+import { generators } from 'openid-client'
+
+import * as Errors from './error'
+
+/**
+ * Generates a PKCE challenge pair where `codeChallenge` is the generated S256 challenge from `codeVerifier`
+ * @param length The length of the code verifier
+ * @returns The generated challenge pair
+ */
+export function generatePkcePair(length = 43): {
```
mantariksh, line 10: thought the cool kids used arrow functions these days (either is fine)

src/sgidClient.ts
mantariksh, line 1: nit: we can name this file `SgidClient.ts` to make it clear that it exports a class

src/constants.ts
```diff
@@ -0,0 +1,9 @@
...
+export const DEFAULT_SGID_CODE_CHALLENGE_METHOD = 'S256'
+export const DEFAULT_SCOPE = 'myinfo.name openid'
+export const SGID_SUPPORTED_FLOWS: ResponseType[] = ['code']
```
mantariksh: hrm "flows" is quite generic, I wonder if we can name this variable to be more specific? e.g. grant types or response types

src/error.ts
```diff
@@ -1,3 +1,9 @@
+// PKCE generatePkcePair and generateCodeVerifier errors
+export const CODE_VERIFIER_LENGTH_ERROR =
+  'The code verifier should have a minimum length of 43 and a maximum length of 128'
```
mantariksh:
```suggestion
  'Code verifier should have a minimum length of 43 and a maximum length of 128'
```

src/error.ts
```diff
@@ -1,3 +1,9 @@
...
+export const PKCE_PAIR_LENGTH_ERROR =
+  'The function generatePkcePair should receive a minimum length of 43 and a maximum length of 128'
```
mantariksh:
```suggestion
  'generatePkcePair should receive a minimum length of 43 and a maximum length of 128'
```

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+     */
+    this.privateKey = this.privateKey.replace(/\\n/gm, '\n')
+  }
+
+  /**
+   * Generates authorization url to redirect end-user to sgID login page.
+   * @param state A string which will be passed back to your application once the end-user
+   * logs in. You should use this to prevent cross-site request forgery attacks (see
+   * https://www.rfc-editor.org/rfc/rfc6749#section-10.12). You can also use this to
+   * track per-request state.
```
mantariksh: can remove the line on CSRF

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+   * @param scope Array or space-separated scopes. 'openid' must be provided as a scope.
+   * Defaults to 'myinfo.nric_number openid'.
```
mantariksh: can update these defaults

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+   * @param nonce Unique nonce for this request. If this param is undefined, a nonce is generated
+   * and returned. To prevent this behaviour, specify null for this param.
+   * @param redirectUri The redirect URI used in the authorization request. If this param is provided,
+   * it will be used over the redirect URI provided in the SgidClient constructor. The redirect URI
```
mantariksh:
```suggestion
   * it will be used instead of the redirect URI provided in the SgidClient constructor. The redirect URI
```

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+   * must be provided in at least the SgidClient constructor or this function.
```
mantariksh: hm think we need to rethink the English here. maybe "If not provided in the constructor, the redirect URI must be provided here."

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+    }
+    return this.sgID.metadata.redirect_uris[0]
+  }
+
+  /**
+   * Exchanges authorization code for access token.
+   * @param code The authorization code received from the authorization server
+   * @param nonce Nonce passed to authorizationUrl for this request. Specify null
+   * if no nonce was passed to authorizationUrl.
+   * @param redirectUri The redirect URI used in the authorization request. Defaults to the one
```
mantariksh: if redirect URI not provided in constructor or here, do we need to throw an error here too, just like in authorizationUrl?

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+   * passed to the SgidClient constructor.
+   * @param codeVerifier The code verifier that was used to generate the code challenge that was passed in `authorizationUrl`
+   * @returns The sub (subject identifier claim) of the user and access token. The subject
```
mantariksh: returns the access token too, no?

src/sgidClient.ts
```diff
@@ -0,0 +1,233 @@
...
+      key: string | undefined
+      data: Record<string, string> | undefined
+    }>(accessToken)
+
+    if (encryptedPayloadKey && data) {
+      const result = await this.decryptPayload(encryptedPayloadKey, data)
+      return { sub, data: result }
+    }
+
+    return { sub, data: {} }
```
mantariksh: I know a while ago I commended jiehao for writing a good [PR description](https://github.com/opengovsg/sgid-client/pull/14) explaining why we return this when there's no key/data, but it would have been even better to leave a comment here to explain. shall we do that here?

.eslintrc
raynerljm: haha yeah i'm aware, i talked to him about it and he said it's okay if we remove it if we decide it's what we want to do as a team :D

README.md
raynerljm: mm yeah great idea!

README.md
raynerljm: wow good catch, i stared at the line for a good 30s before i realized the typo

README.md
raynerljm: yup okay i agree with that too

README.md
raynerljm: oh yes thank you

README.md
raynerljm: yeah that completely makes sense!

README.md
raynerljm: mm yeah that makes sense

README.md
raynerljm: image

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.7.1

Seems like nonce is encouraged but not mandatory (just like state) because of PKCE protection. i think let's still encourage it nonetheless. i'll change the phrasing to be more similar to state.

README.md
raynerljm: yup will change it

README.md
raynerljm: i've added a comment for nonce in `callback` to specify that it should be the nonce generated from `authorizationUrl`. feel free to comment if you think it can be clearer!

README.md
raynerljm: yup will do so

src/generators.ts
raynerljm: haha i like arrow functions, but defining function return types explicitly is recommended in library code and i feel that it looks clearer with `function` instead of arrow functions
[@typescript-eslint/explicit-module-boundary-types](https://github.com/typescript-eslint/typescript-eslint/blob/v4.25.0/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md)

README.md
raynerljm: i've added some descriptions about what they are and how to use it! feel free to comment if you feel it can be improved further

README.md
raynerljm: i've tweaked the README so that
- all the code snippet inline comments are moved to the function type signature instead (so that readers don't have to read both the function type signature explanation + code snippet explanation)
- added return type for the function signature as well

src/sgidClient.ts
raynerljm: yup that makes perfect sense

raynerljm: note i had to make a commit to rename it wrongly (i.e. from `sgidClient.ts` to `SsgidClient.ts`) before renaming it correctly to (`SgidClient.ts`). this is because from my experience, git diffs don't treat case-sensitive changes well and causes a mismatch between the local environment + github

src/error.ts
raynerljm: will make the change

src/error.ts
raynerljm: will make the change

raynerljm: will make the change

src/sgidClient.ts
raynerljm: yup i think we need to throw an error because the token endpoint expects a redirect URI. the function `getFirstRedirectUri` is used in both `authorizationUrl` and `callback` and is the function that will throw the error.

src/sgidClient.ts
raynerljm: sounds good!

src/sgidClient.ts
raynerljm: sounds good!

src/sgidClient.ts
raynerljm: good catch! updated

src/sgidClient.ts
raynerljm: okay removed

src/sgidClient.ts
raynerljm: yup it returns the access token too - the full comment is `The sub (subject identifier claim) of the user and access token. The subject identifier claim is the end-user's unique ID.`

do you think it's sufficient or can we make it clearer?

src/sgidClient.ts
raynerljm: yeah that's a great idea! will add a comment

src/sgidClient.ts
mantariksh: oops lol clearly I misread, my bad. current version makes sense!

README.md
```diff
@@ -18,53 +14,113 @@ npm i @opengovsg/sgid-client
...
+const { codeChallenge, codeVerifier } = generatePkcePair()
+```
+
 ### Get Authorization URL
 
-`client.authorizationUrl(state, scope, [nonce], [redirectUri])`
+`client.authorizationUrl(parameters)`
+
+- parameters: `<Object>`
+  - codeChallenge: `<string>` Code challenge generated from the code verifier.
```
mantariksh: user doesn't need to know specifics of how the code challenge and verifier are related. from their pov they just call a magic function :)
```suggestion
  - codeChallenge: `<string>` Code challenge generated from `generatePkcePair`
```

README.md
```diff
@@ -18,53 +14,113 @@ npm i @opengovsg/sgid-client
...
+  - state: `<string>` (Optional) Used to track application state which will be passed back to you in query params via the redirect URI after login.
+  - scope: `<string> | <string[]>` (Optional) Scopes being requested. Can be provided as a string array or a space-concatenated string. **Default**: `myinfo.name openid`
+  - nonce: `<string> | <null>` (Optional) Random, unique value to associate a user-session with an ID Token and to mitigate replay attacks. Set as `null` to omit the nonce. **Default**: Randomly generates a nonce if unspecified or set as `undefined`
+  - redirectUri: `<string>` (Optional) Overrides redirect URI initially provided in constructor. **Default**: Utilizes the `redirectUri` provided in the constructor.
+- Returns: `<Object>`
+  - url: `<string>` Generated authorization url.
+  - nonce: `<string> | <undefined>` Provided nonce, randomly generated nonce, or `undefined` (based on nonce input).
```
mantariksh: just as with code verifier, we can let them know that the nonce should be stored in the user's session so it can be retrieved in `callback` later

src/sgidClient.ts
mantariksh: ah ok thanks for explaining!

test/SgidClient.spec.ts
```diff
@@ -58,18 +68,21 @@ describe('SgidClient', () => {
...
-        apiVersion: MOCK_API_VERSION,
       })
 
       expect(pkcs8Client).toBeDefined()
     })
   })
 
   describe('authorizationUrl', () => {
-    it('should generate authorisation URL correctly when only state is provided', () => {
+    it('should generate authorisation URL correctly when state and codeChallenge is provided', () => {
```
mantariksh:
```suggestion
    it('should generate authorisation URL correctly when state and codeChallenge are provided', () => {
```

test/SgidClient.spec.ts
```diff
@@ -305,28 +376,105 @@ describe('SgidClient', () => {
...
+
+      expect(pkcePair.codeVerifier.length).toBe(43)
+      expect(pkcePair.codeChallenge.length).toBe(43)
+
+      expect(pkcePair.codeChallenge).toMatch(codeVerifierAndChallengePattern)
+      expect(pkcePair.codeVerifier).toMatch(codeVerifierAndChallengePattern)
+    })
+
+    it('should generate a PKCE pair of specified length when length between 43 (inclusive) and 128 (inclusive) is provided', () => {
+      for (let length = 43; length <= 128; length++) {
```
mantariksh, line 422: omG

mantariksh: how much time does this add to the tests? if it's cheap then no issue, but if this adding more than say 30s to the tests, I'm not sure it's worth it. we can just sample 3-4 sizes to test instead of testing every single number

test/SgidClient.spec.ts
```diff
@@ -305,28 +376,105 @@ describe('SgidClient', () => {
...
+
+  describe('generatePkcePair', () => {
+    it('should generate a PKCE pair when no length is provided', () => {
+      const pkcePair = generatePkcePair()
+
+      expect(pkcePair.codeVerifier.length).toBe(43)
+      expect(pkcePair.codeChallenge.length).toBe(43)
+
+      expect(pkcePair.codeChallenge).toMatch(codeVerifierAndChallengePattern)
+      expect(pkcePair.codeVerifier).toMatch(codeVerifierAndChallengePattern)
```
mantariksh: should we also be checking that the challenge is the correct hash of the verifier?

test/SgidClient.spec.ts
raynerljm: the entire test suite takes like ~1 second so i think it's fine :D

README.md
raynerljm: mm makes sense

test/SgidClient.spec.ts
raynerljm: ahh good catch!

test/SgidClient.spec.ts
raynerljm: ahh yes - i forgot that i omitted it because of the assumption that the underlying functions work. great catch thanks!

README.md
raynerljm: makes sense!

src/constants.ts
raynerljm: renamed to `SGID_SUPPORTED_GRANT_TYPES`

src/sgidClient.ts
raynerljm: @kwajiehao brought up a good point that since the `API_VERSION` value is hardcoded (and no longer a constructor param), there's not really any point to do a dynamic check like that because we always know what `API_VERSION` will be

mantariksh (approved):
gtml gooks tood mo le

release plan should minimally test against sgID production rather than a local sgID server, and ideally test in a production environment rather than a local environment

mantariksh (approved):
gtml gooks tood mo le

release plan should minimally test against sgID production rather than a local sgID server, and ideally test in a production environment rather than a local environment

raynerljm:
> gtml gooks tood mo le
> 
> release plan should minimally test against sgID production rather than a local sgID server, and ideally test in a production environment rather than a local environment

yup totally agree - here is a [draft of the release plan](https://www.notion.so/opengov/TypeScript-SDK-v2-0-Release-Plan-2dc92c1c03b64d5cb821161edfd65131?pvs=4) which involves how the release will be done and testing in local + prod + a mix

src/sgidClient.ts
kschiew: Just for

> note i had to make a commit to rename it wrongly (i.e. from `sgidClient.ts` to `SsgidClient.ts`) before renaming it correctly to (`SgidClient.ts`). this is because from my experience, git diffs don't treat case-sensitive changes well and causes a mismatch between the local environment + github

Just out of curiosity, what is the inconvenience caused from case-sensitive changes?

kschiew (approved):
Did one last sanity check , lgtm!

Issue #38, opened 2023-04-26
Refactor sgid-client


raynerljm:
Refactor the SDK so that 
- it is more organized
- exports are standardized (and more straightforward to use from the POV of a developer)
- ~~unit tests are split into smaller test suites with their own responsibility~~ (removed from the scope of the PR)

Closed in this PR https://github.com/opengovsg/sgid-client/pull/46

raynerljm:
Closed as it is completed

PR #39, opened 2023-04-26
(OLD) refactor: Refactor and reorganize files and exports

raynerljm:
(OLD): Some of these changes were implemented in #35 . I will make a new PR specifically for test-related changes subsequently

## Problem

General refactoring of files and unit tests

Closes issue #38 https://github.com/orgs/datagovsg/projects/43/views/1?pane=issue&itemId=26650732

Depends on #35 

## Solution

Refactor the 
- `SgidClient` class into its own `sgidClient.ts` file 
- unit tests (split the testing of the functions into their own files)
- constants into its own `constants.ts` file
- generators (e.g. `generatePkcePair`) into `generators.ts` file)
- move `decryptPayload` function into the `utils.ts` file and expose it publicly
- use the `index.ts` just as a tunnel to export the main exposed functions (i.e. `sgidClient.ts`, `generators.ts`, and `utils.ts`

## Tests

All unit tests are still passing after the refactor

image


src/sgidClient.ts
```diff
@@ -0,0 +1,196 @@
...
+   * @returns
+   */
+  authorizationUrl({
+    state,
+    scope = DEFAULT_SCOPE,
+    nonce = generators.nonce(),
+    redirectUri = this.getFirstRedirectUri(),
+    codeChallenge,
+  }: AuthorizationUrlParams): AuthorizationUrlReturn {
+    if (this.apiVersion !== 2) {
```
kwajiehao, line 82: See comments on #35 for this

test/authorizationUrl.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    })
+
+    const actual = new URL(url)
+    // Count number of search params
+    let actualNumSearchParams = 0
+    actual.searchParams.forEach(() => actualNumSearchParams++)
+    const expected = new URL(MOCK_AUTH_ENDPOINT_V2)
+    expect(actual.host).toBe(expected.host)
+    expect(actual.pathname).toBe(expected.pathname)
+    expect(actual.searchParams.get('client_id')).toBe(MOCK_CLIENT_ID)
```
kwajiehao, line 111: nit: currently, there are many repetitive lines of `expect`s across the tests. instead of writing attribute by attribute, what do you think of using either a string or object to compare (especially considering most values are checked against the default values)? 

this allows you to construct a base, expected search params object which you can modify as needed to get the expected result

test/constructor.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      redirectUris: [MOCK_REDIRECT_URI],
+      hostname: MOCK_HOSTNAME,
+    })
+  })
+
+  it('should initialise correctly when a PKCS1 private key is provided', () => {
+    expect(client).toBeDefined()
+  })
+
+  it('should initialise correctly when a PKCS8 private key is provided', () => {
```
kwajiehao, line 30: we can add a test case that checks for whether new lines are stripped from the private key

src/util.ts
```diff
@@ -9,7 +10,49 @@ export function convertPkcs1ToPkcs8(pkcs1: string): string {
...
- * Regex pattern that the code verifier and code challenge in the PKCE flow should match according to the PKCE RFC
- * https://www.rfc-editor.org/rfc/rfc7636
+ * Decrypts the payload received from the `userinfo` function
+ * @param privateKey sgID client private key to decrypt the encrypted payload key
+ * @param encryptedPayloadKey Encrypted payload key that when decrypted, is used to decrypt the encrypted data
+ * @param data Encrypted data
+ * @returns Decrypted data and unique sgID
  */
-export const codeVerifierAndChallengePattern = /^[A-Za-z\d\-._~]{43,128}$/
+export async function decryptPayload(
```
kwajiehao, line 19: Can we also have unit tests for `decryptPayload`?

test/authorizationUrl.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    client = new SgidClient({
+      clientId: MOCK_CLIENT_ID,
+      clientSecret: MOCK_CLIENT_SECRET,
+      privateKey: MOCK_CLIENT_PRIVATE_KEY,
+      redirectUris: [MOCK_REDIRECT_URI],
+      hostname: MOCK_HOSTNAME,
+    })
+  })
+
+  it('should generate authorisation URL correctly when state and codeChallenge is provided', () => {
```
kwajiehao, line 30: Actually, we don't need `state` to be mandatory since we are mandating PKCE

test/authorizationUrl.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    expect(actual.searchParams.get('state')).toBe(mockState)
+    expect(actual.searchParams.get('code_challenge')).toBe(mockCodeChallenge)
+    expect(actual.searchParams.get('code_challenge_method')).toBe(
+      DEFAULT_SGID_CODE_CHALLENGE_METHOD,
+    )
+    // Client ID, scope, response_type, redirect_uri, nonce, state, code_challenge, code_challenge_method
+    expect(actualNumSearchParams).toBe(8)
+  })
+
+  it('should throw when no redirectUri is provided', () => {
```
kwajiehao, line 226: We should have a negative test case for when no code challenge is provided

test/callback.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      codeVerifier: MOCK_CODE_VERIFIER,
+    })
+
+    expect(sub).toBe(MOCK_SUB)
+    expect(accessToken).toBe(MOCK_ACCESS_TOKEN)
+  })
+
+  it('should throw when no code verifier is provided', async () => {
+    await expect(client.callback({ code: MOCK_AUTH_CODE })).rejects.toThrow(
+      "Code verifier must be provided in 'callback' when using apiVersion 2",
```
kwajiehao, line 42: See comments in #35 on api version

Also, ultimately both PRs should be rebased against #27 / #27 should be merged into both PRs so that all error messages are in a central file. We can set up a session to go through the rebase together

kwajiehao (commented):
Left comments!

test/callback.spec.ts
raynerljm: yup i definitely agree on the rebase with #27

src/constants.ts
```diff
@@ -1,9 +1,30 @@
...
 export const API_VERSION = 2
+
+/**
+ * Testing
+ */
+/**
+ * Regex pattern that the code verifier and code challenge in the PKCE flow should match according to the PKCE RFC
+ * https://www.rfc-editor.org/rfc/rfc7636
+ */
+export const codeVerifierAndChallengePattern = /^[A-Za-z\d\-._~]{43,128}$/
```
mantariksh, line 28: is this used purely in test code? if so, we should put it together with the test code instead of bundling it together with SDK code. it's good to be deliberate about what we export, as we have to support it forever or make a breaking change to remove it

raynerljm: mm that's a very good point - shifted into the test code in this [PR ](https://github.com/opengovsg/sgid-client/pull/35)

raynerljm:
Closed because most of these changes were implemented in https://github.com/opengovsg/sgid-client/pull/35 .

PR #42, opened 2023-05-05
chore: upgrade all linting packages

mantariksh:
## Problem

All our linting packages are behind by at least 1 major version. Upgrading makes it easier to set up linting in subfolders (for adding example backends), and also makes it easier to upgrade to future major versions.

## Solution

- Upgrade all linting packages
- Make a test commit (5e40c4d9b22d2aa0464d489a87b2c4d267d11c35) with changes to all ts files, and check that linting runs fine on commit
- Revert above commit


PrawiraGenestonlia (approved):
lgtm!

PR #43, opened 2023-05-05
feat: example app

mantariksh:
TODO: add README


examples/express/README.md
```diff
@@ -0,0 +1,42 @@
+# ExpressJS example sgID app
+
+The example application code is in `index.ts`. You can copy this code to bootstrap your sgID client application, and run this app locally to understand how this SDK helps you interact with the sgID server.
+
+## Running this example app locally
+
+### Prerequisites
+
+Register a new client at the [sgID developer portal](https://developer.id.gov.sg). Feel free to register a test client; there is no limit to the number of clients you can create.
```
raynerljm, line 9: I think it would be good to:
1. give more detail (e.g. what `redirect_uri` to register)
2. link to the [registration page ](https://jie-hao-kwa.gitbook.io/sgid-v2-beta/~/changes/75HA2ZSHqQMuHPNECzfs/introduction/getting-started/registration)on our docs for more details

examples/express/helpers.ts
```diff
@@ -0,0 +1,22 @@
+import fs from 'fs'
+import fetch from 'node-fetch'
+import extract from 'extract-zip'
+import path from 'path'
+
+const SOURCE_URL = 'https://d2m9b44i9bvctd.cloudfront.net/dist.zip'
+
+export const fetchStaticFiles = async (): Promise<void> => {
```
raynerljm, line 8: (please correct me if i'm wrong) from what i understand, making the express app fetch the frontend files and serve them like this means this is a [traditional web server /SSR](https://app.gitbook.com/s/pBW92htBuXTrDYoKovQ2/~/changes/1/learn-the-basics/integration-patterns/web-server-and-ssr-front-end/~/comments/2PeVjMn4dKMUr84kiS2H) example

would creating a new example/ explaining how to modify this Express server to be a [BFF](https://app.gitbook.com/s/pBW92htBuXTrDYoKovQ2/~/changes/1/learn-the-basics/integration-patterns/back-end-for-spa-front-end-bff/~/comments/wQwRwhbQ4f0dBmuMYJvv) instead be something we want to consider doing?

raynerljm: in the context of LifeHack, from what i know from my peers + what i've seen from hackathon submissions, i think students nowadays are a lot more familiar with a SPA Frontend + Backend than a MVC style web server 

if our immediate goal is to make sgID as easy to onboard as possible for LifeHack, creating an example for the BFF integration pattern might be quite important

examples/express/index.ts
```diff
@@ -0,0 +1,135 @@
...
+import SgidClient from '@opengovsg/sgid-client'
+import * as dotenv from 'dotenv'
+import crypto from 'crypto'
+import cookieParser from 'cookie-parser'
+import { fetchStaticFiles } from './helpers'
+import open from 'open'
+
+dotenv.config()
+
+const PORT = 5000
```
raynerljm, line 11: port 5000 has some [issues](https://developer.apple.com/forums/thread/682332) on macs (@kschiew and i faced this issue). i think we should move to a less problematic port haha

examples/express/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Scopes that all sgID relying parties can access by default
+    ['openid', 'myinfo.name'],
+  )
+  sessionData[sessionId] = {
+    state,
+    nonce,
+  }
+  return res
+    .cookie(SESSION_COOKIE_NAME, sessionId, SESSION_COOKIE_OPTIONS)
+    .json({ url })
```
raynerljm, line 68: is there a reason that we are returning the `url` to the FE and letting the FE redirect instead of directly redirecting from this endpoint?
e.g.
```
res.redirect(url)
```

examples/express/index.ts
```diff
@@ -0,0 +1,135 @@
...
+      accessToken?: string
+    }
+  | undefined
+>
+
+/**
+ * In-memory store for session data.
+ * In a real application, this would be a database.
+ */
+const sessionData: SessionData = {}
```
raynerljm, line 44: nit: have we considered using a `Map` instead of an object? in a demo app like this it doesn't really matter but using a `Map`:
1. is more suited for a use case like this than object (due to [performance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map))
2. we don't need to explictly type the Record value as undefined
```suggestion
type Session = {
      nonce?: string
      state: URLSearchParams
      accessToken?: string
}

const sessionData = new Map<string, Session>()

const session = sessionData.get("123xyz")
//^? typeof session === Session | undefined
```

however, `Map` is definitely not as commonly used as objects so i'm afraid that making this change might make it less understandable for some readers

examples/express/helpers.ts
raynerljm: Resolved: Discussed offline - it's not a traditional web server even though the SPA is served from the same server (SPA is CSR, traditional web server is SSR)

examples/express/index.ts
kwajiehao: The reason for this is that the redirect request will be blocked by CORS because `api.id.gov.sg` does not have the proper CORS headers set. This happens because this is a response from the initial request by the browser

PR #46, opened 2023-05-15
feat: PKCE (redo)

raynerljm:
> **Note**: For the full list of reviews/comments, please refer to this [PR](https://github.com/opengovsg/sgid-client/pull/35).(This PR was re-created to squash merge instead of merge)

## Problem

Supporting the [[Proof Key for Code Exchange (PKCE)](https://www.rfc-editor.org/rfc/rfc7636)](https://www.rfc-editor.org/rfc/rfc7636) extension to OAuth 2.0 is a current best practice for preventing authorisation code interception attacks in native apps. As per [[RFC 8252 (OAuth 2.0 for Native Apps)](https://www.rfc-editor.org/rfc/rfc8252)](https://www.rfc-editor.org/rfc/rfc8252):

> 

### High-level principles

As this is a relatively large PR which involves breaking changes, I would like to highlight a few high-level principles that were kept in mind during development and which should be achieved through the code changes.

1. The SDK should be dummy

This is in line with what we have been doing so far where we would like to keep the SDK dummy. This is so that if we would like to make any changes (e.g. with validation), we are able to modify our servers directly instead of updating our SDK (as not all RPs will update their SDKs).

This is achieved by not validating the ABNF of the codeChallenge and codeVerifier received in `authorizationUrl` and `callback` respectively. (Especially because validation bugs are easier to commit and harder to rectify)

2. We should be mindful of the exports we are providing

As we are maintaining a public SDK for people to use, we need to be very aware of what exactly we are exposing for developers to import. If we export function `a`, we have to continue maintaining `a` as an export in future versions as developers might be relying on them.

This is achieved by refactoring the `SgidClient` class into it's own file and keeping `index.ts` solely as a tunnel for exports. Additionally, we are now exporting the SDK types so developers with TypeScript can make use of them.

3. We should be mindful of the SDK function signature

With this PR, the API of the SDK is changing (i.e.  the function signature is different). Specifically, the functions now take in an <Object> instead of sequential parameters.

This is because as the list of parameters gets longer this (i) makes it harder to remember which input corresponds to which parameter and (ii) makes it very confusing when there are optional parameters - users would have to enter `undefined` for the optional parameters they do not want to fill in just to reach the parameter they have to fill in.

**Features**:

PKCE related changes
- Added the `generatePkcePair`, `generateCodeVerifier` and `generateCodeChallenge` functions based on `openid-client` `generator` functions in the `generators.ts` file
- Changed the `authorizationUrl` function params to take in an object
    - Added the `codeChallenge` ~~and `codeChallengeMethod`~~ parameter
- Changed the `callback` function params to take in an object
    - Added the `codeVerifier` parameter
- Added relevant tests for new functions
    - Added tests for `generatePkcePair`, `generateCodeVerifier` and `generateCodeChallenge` functions
    - Updated tests for `authorizationUrl` and `callback`
    
SDK improvement related changes
- Changed the `userinfo` function params to take in an object
- Changed default scope from `openid myinfo.nric_number` to `openid myinfo.name` 
- Refactored main `SgidClient` class into its own file `sgidClient.ts` and extracted types into `types.ts`
- Exported `sgidClient.ts`, `types.ts`, `generators.ts` and `util.ts` through `index.ts` (to be explicit about what we want exported)
- Removed `apiVersion` as a parameter for the constructor and use a hardcoded `API_VERSION` constant
- Updated the issuer to have the `v2` appended (in the constructor)
- ~~Turn off~~ Remove `eslint-no-throw-sync-func` ESLint rule 

README related changes
- Added description + example usage for generatePkcePair
- Added function signature (parameters + return type) indicating the type, whether it's optional, the default value, and a brief explanation
- Removed the code snippet inline comments to the function signature

Non-implemented changes
~~- Changing the constructor param from `redirectUri` to `redirectUris`~~

## Tests
- Added test suites for `generatePkcePair`, `generateCodeVerifier`, and `generateCodeChallenge`
- Updated test cases for `authorizationUrl` and `callback` functions
- Tested E2E locally following [this flow](https://www.notion.so/opengov/Testing-flow-from-Test-App-thru-Node-js-SDK-to-Local-server-8e9c301e7e36403d95fcfc02f9dedf8b?pvs=4)

**Verified that all tests are passing**
image


PrawiraGenestonlia (approved):
lgtm!

PR #47, opened 2023-05-15
feat: add Next.js CSR example

raynerljm:
## Problem

Our SDK currently lacks example apps demonstrating how to use our SDK. This introduces two problems
1. RPs do not have example apps following good practices to refer to
2. We are not able to easily test SDK changes with an app (we need to do an `npm link`)

Closes #384

## Solution

Added an example Next.js app using CSR and API routes. It is hosted [here](https://sgid-nextjs-csr-demo.vercel.app/).

**Features**:
- Primarily communicates with `sgid-server` via the `src/pages/api` routes
- Utilizes an in-memory store `src/lib/store.js` to store Session data (e.g. state, nonce, codeVerifier, etc.)

## Tests

// TODO: ADD recording


examples/.DS_Store
kwajiehao, line 1: Same comment as in #49, remove all `.DS_Store` files and add to `.gitignore`

examples/nextjs-csr/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
+
+## Getting Started
+
+Before you can run the development server, you will have to register your client on the sgID [developer portal](https://developer.id.gov.sg/).
+
+> For more information about sgID, please visit the [developer documentation](https://docs.id.gov.sg/).
+
+Copy the `.env.example` file, rename it to `.env`, and fill in your credentials obtained during registration.
```
kwajiehao, line 13: Missing `.env.example` file

examples/nextjs-csr/src/lib/sgidClient.ts
```diff
@@ -0,0 +1,13 @@
+import { SgidClient } from "@opengovsg/sgid-client";
+
+const sgidClient = new SgidClient({
+  clientId: String(process.env.SGID_CLIENT_ID) || "",
+  clientSecret: String(process.env.SGID_CLIENT_SECRET) || "",
+  privateKey: String(process.env.SGID_PRIVATE_KEY).replace(/\\n/g, "\n") || "",
+  redirectUri:
```
kwajiehao: What is the reason for the difference in implementation here and in #49, where the redirect URI for local dev was commented out?

examples/nextjs-csr/src/pages/api/callback.ts
```diff
@@ -0,0 +1,50 @@
...
+    nonce,
+    codeVerifier,
+  });
+
+  const newSession = {
+    ...session,
+    accessToken,
+  };
+
+  store.set(sessionId, newSession);
```
kwajiehao: Should we clear the old session once the new session is created?

examples/nextjs-csr/src/pages/api/callback.ts
```diff
@@ -0,0 +1,50 @@
...
+  } else if (
+    !session.codeVerifier ||
+    typeof session.codeVerifier !== "string"
+  ) {
+    return res.status(400).send("Code verifier not found");
+  }
+
+  const { nonce, codeVerifier } = session;
+
+  const { accessToken } = await sgidClient.callback({
```
kwajiehao: At the end of this function, we log the user in. Theoretically, this is correct, because behind the scenes, the `openid-client` library validates the ID token, which is the actual proof of authentication. Not sure if we need to add any comments here to clarify, but I just wanted to put this here for my own clarity

examples/nextjs-csr/src/pages/api/userinfo.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  if (typeof sessionId !== "string") {
+    return res.status(400).send("Session ID not found in browser cookies");
+  }
+
+  const session = store.get(sessionId);
+
+  if (!session) {
+    return res.status(400).send("Session not found");
+  } else if (!session.accessToken || typeof session.accessToken !== "string") {
```
kwajiehao: nit: for code cleanliness, perhaps the checks for `accessToken` can be done after we destructure it:

```
const { accessToken } = session;
if (!accessToken || typeof accessToken !== "string") {
    return res.status(400).send("Access token not in session");
}
```

examples/nextjs-csr/src/pages/logged-in.tsx
```diff
@@ -0,0 +1,81 @@
...
+  state?: string;
+};
+
+const LoggedIn = () => {
+  const [data, setData] = useState<UserInfoRes | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState("");
+
+  useEffect(() => {
+    const getUserInfo = async () => {
```
kwajiehao, line 16: Currently, the `userinfo` step and callback handling steps are performed separately, which seems to be different from what we agreed upon in the backend for frontend flow (https://app.gitbook.com/s/pBW92htBuXTrDYoKovQ2/~/changes/9/learn-the-basics/integration-patterns/backend-for-single-page-app-spa-frontend-bff). Could you elaborate on the rationale for splitting them?

kwajiehao (commented):
Left comments! happy to discuss them over a call as well

examples/nextjs-csr/src/pages/logged-in.tsx
raynerljm: hmm i was under the impression that we were planning to perform them separately because of the [lucid chart diagram](https://lucid.app/lucidchart/477b82f6-19d2-4010-a8af-06e12688bd3b/edit?invitationId=inv_21f0e528-8e2a-40b7-8546-6ebbc3f1f617&page=0_0#) for the demo app that we did + Anty's Express demo app also separates them. 

I'll check in with Wira as he drew up the diagram

raynerljm: Discussed with Wira - separating the userinfo and callback steps are more suited for session-based authentication but the diagram in the GitBook bff page is more generic and does not describe a userinfo endpoint. this is so that the RPs can decide whether they would like to employ session-based or token-based authentication for their app. 

do you think it's important for the app to be in-sync with the diagram for client authentication? (currently all our CSR examples split them)

examples/.DS_Store
raynerljm: yup will do that

examples/nextjs-csr/README.md
raynerljm: okay thanks will add it, for some reason it didn't get copied over

examples/nextjs-csr/src/lib/sgidClient.ts
raynerljm: the Next.js SSR example is still WIP and hasn't been updated - will update to follow this instead

examples/nextjs-csr/src/pages/api/callback.ts
raynerljm: oh it's not actually a new session, the key is the same session ID. it's just that the session is updated with more data. I'll rename it to `updatedSession` make it clearer

examples/nextjs-csr/src/pages/api/callback.ts
raynerljm: yeah i think it would be helpful to add that as a comment as well

examples/nextjs-csr/src/pages/api/userinfo.ts
raynerljm: good idea!

raynerljm:
@kwajiehao also made the following changes based on comments from the [Next.js SSR](https://github.com/opengovsg/sgid-client/pull/49) PR.
- Added the scopes and redirect URI to register in the README
- Let `start:dev` script do both the build and running

examples/nextjs-csr/src/pages/api/callback.ts
kwajiehao: My bad, misread your code!

examples/nextjs-csr/src/pages/logged-in.tsx
kwajiehao: gotcha, thank you for clarifying! 

i don't think it's a big deal because both flows are secure, and I think it's fine to go with the flow with the split routes because that is what Anty and Wira have already agreed on. 

re: the inconsistency between the diagram and the code example, the diagram does refer to a userinfo endpoint. In an ideal world the example flow would be consistent with the diagram but I think the differences are small, and we can always fix the diagram later so let's just roll with it for now. What do you think?

PrawiraGenestonlia: The differences in the diagram are on purpose because I didn't want to suggest a client authentication mechanism within the flow (in our example, we are using session-in-cookies based authentication and hence, the routes are different. I told @raynerljm to proceed with the example because we can always update the diagram whenever necessary.

examples/nextjs-csr/src/lib/sgidClient.ts
raynerljm: SSR example has been updated [here](https://github.com/opengovsg/sgid-client/pull/49)

examples/nextjs-csr/src/pages/logged-in.tsx
raynerljm: yup - i completely agree that we can always fix the diagram in the future. let's proceed with this first :D thanks for the discussion!

kwajiehao (approved):
lgtm

PR #48, opened 2023-05-15
feat: check that userinfo sub matches ID token

mantariksh:
## Problem

[OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse) states that the RP must verify that the `sub` returned by the userinfo endpoint matches the `sub` in the ID token. Since this is a requirement for all RPs, the SDK should implement this check.

## Solution

Change the API of the `userinfo` method to include the `sub`, and check that the `sub` passed matches the `sub` returned by the userinfo endpoint.

### DIFF
--- a/src/SgidClient.ts
+++ b/src/SgidClient.ts
@@ -163,17 +163,22 @@ export class SgidClient {
 
   /**
    * Retrieves verified user info and decrypts it with your private key.
-   * @param accessToken The access token returned in the callback function
-   * @returns The sub of the end-user and the end-user's verified data
+   * @param sub The sub returned from the callback function
+   * @param accessToken The access token returned from the callback function
+   * @returns The sub of the end-user and the end-user's verified data. The sub
+   * returned is the same as the one passed in the params.
    */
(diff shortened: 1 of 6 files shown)


src/SgidClient.ts
```diff
@@ -163,17 +163,22 @@ export class SgidClient {
 
   /**
    * Retrieves verified user info and decrypts it with your private key.
-   * @param accessToken The access token returned in the callback function
-   * @returns The sub of the end-user and the end-user's verified data
+   * @param sub The sub returned from the callback function
+   * @param accessToken The access token returned from the callback function
+   * @returns The sub of the end-user and the end-user's verified data. The sub
+   * returned is the same as the one passed in the params.
```
raynerljm, line 169: good idea! i think this would be helpful for some people

raynerljm (approved):
lgtm! good job on adding all the tests and documentation updates so quickly 🌮

kwajiehao (approved):
lgtm

PR #49, opened 2023-05-15
feat: add Nextjs SSR example

raynerljm:
## Problem

Our SDK currently lacks example apps demonstrating how to use our SDK. This introduces two problems
1. RPs do not have example apps following good practices to refer to
2. We are not able to easily test SDK changes with an app (we need to do an `npm link`)

Closes #384

## Solution

Added an example Next.js app using SSR and server components (Next.js 13.4 feature). It is hosted [here](https://sgid-nextjs-ssr-demo.vercel.app/).

**Features**:
- Utilizes an in-memory store `src/lib/store.js` to store Session data (e.g. state, nonce, codeVerifier, etc.)

## Tests

// TODO: ADD recording


examples/nextjs-ssr/README.md
```diff
@@ -0,0 +1,21 @@
...
+> For more information about sgID, please visit the [developer documentation](https://docs.id.gov.sg/).
+
+Copy the `.env.example` file, rename it to `.env`, and fill in your credentials obtained during registration.
+
+Replace the `redirectUri` in `src/lib/sgidClient.ts` with `http://localhost:3000/success`
+
+Then, run the server by running:
+
+```bash
+npm run build && npm run start
```
kwajiehao: nit: create a new script that includes this command so it's easier to run

.eslintrc
```diff
@@ -13,6 +13,7 @@
   "parserOptions": {
     "ecmaVersion": 2018
   },
+  "ignorePatterns": ["examples/**/*"],
```
kwajiehao: just curious, what's the rationale for ignoring the files in examples?

examples/.DS_Store
kwajiehao, line 1: Can we remove this file and add `.DS_Store` to our `.gitignore` file?

examples/nextjs-ssr/src/.DS_Store
kwajiehao, line 1: Again, remove this file

examples/nextjs-ssr/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
+
+## Getting Started
+
+Before you can run the development server, you will have to register your client on the sgID [developer portal](https://developer.id.gov.sg/).
+
+> For more information about sgID, please visit the [developer documentation](https://docs.id.gov.sg/).
+
+Copy the `.env.example` file, rename it to `.env`, and fill in your credentials obtained during registration.
```
kwajiehao, line 13: There is no `.env.example` file

kwajiehao: Related: do we want to use `dotenv`? I'm thinking it'll make things easier for them but happy to hear your thoughts

examples/nextjs-ssr/src/app/success/page.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { sgidClient } from "@/lib/sgidClient";
+import { cookies } from "next/headers";
+
+const getAndStoreUserInfo = async (code: string, sessionId: string) => {
+  const session = store.get(sessionId);
+
+  // Exchange auth code for access token
+  const { accessToken } = await sgidClient.callback(code, session?.nonce);
+
+  // Request user info with acecss token
```
kwajiehao:
```suggestion
  // Request user info with access token
```

examples/nextjs-ssr/src/app/api/auth-url/route.ts
```diff
@@ -0,0 +1,29 @@
...
+  // Generate new session ID
+  const sessionId = uuidv4();
+
+  // Retrieve state form query params
+  const state = new URL(req.url).searchParams.get("state");
+
+  // Generate authorization url
+  const { url, nonce } = sgidClient.authorizationUrl(
+    String(state),
+    "openid myinfo.name"
```
kwajiehao: This assumes the scopes set by the user, maybe we can keep it to just `openid` first so that it definitely works (as opposed to getting a mismatched scopes error)? What do you think?

examples/nextjs-ssr/src/app/api/auth-url/route.ts
```diff
@@ -0,0 +1,29 @@
...
+import { store } from "@/lib/store";
+import { sgidClient } from "@/lib/sgidClient";
+import { v4 as uuidv4 } from "uuid";
+
+export function GET(req: NextRequest) {
+  // Generate new session ID
+  const sessionId = uuidv4();
+
+  // Retrieve state form query params
+  const state = new URL(req.url).searchParams.get("state");
```
kwajiehao: What's the reason for generating the state as a param instead of in the route itself?

Also I guess if we switch to PKCE we don't need to use state anymore

examples/nextjs-ssr/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Replace the `redirectUri` in `src/lib/sgidClient.ts` with `http://localhost:3000/success`
```
kwajiehao: We should add a note that they need to add this redirect URL to their client registration

kwajiehao (commented):
Left some comments, but otherwise looks good to me!

I have 2 more points I wanted to talk about:

## 1. PKCE 
What is the plan for this? I think it might be useful to include PKCE in this example, keen to hear your thoughts on how we planned to roll that out

## 2. Error-handling
I know we are using an in-memory data store, which makes sense. But currently if you complete the login flow, and then refresh the page, you get this error page:

Screenshot 2023-05-16 at 2 17 43 PM

I'm wondering if it would be more useful if we can render an error page instead with some information on why this is not showing / some other kind of error message. 

What do you think?

examples/nextjs-ssr/README.md
raynerljm: hm what do you mean by use `dotenv`? Next.js support environment variables out of the box so I don't really understand why we would need to use `dotenv`. could you help to clarify?

examples/nextjs-ssr/README.md
raynerljm: yup good idea!

.eslintrc
raynerljm: currently, if the file is not ignored, this error happens when trying to make a commit (not just for the `sgidClient` file but for the entire directory)
image

as each example app has its own ESLint file as well (generated by Next.js), I was thinking it would be easier to just omit the ESLint checking for the examples - otherwise we would have them into the root `tsconfig.json`

I also do see the merit of standardizing linting and formatting across all examples though - what do you think?

examples/.DS_Store
raynerljm: yup okay!

examples/nextjs-ssr/src/app/api/auth-url/route.ts
raynerljm: for the purposes of the demo, i think it would be helpful to retrieve some myinfo data (i.e. `myinfo.name`) as well as `openid` as a proof of concept for the developers. i think we can mitigate this problem by specifying in the README what scopes to set for this example (as well as the redirect URI to specify)

what do you think? I've already added it to the README so you can take a look

examples/nextjs-ssr/README.md
raynerljm: yup added!

examples/nextjs-ssr/src/app/api/auth-url/route.ts
raynerljm: i don't really understand what you mean by 'generating the state as a param instead of in the route itself'. do you mean why is the state being extracted from the `req` the way it is?

additionally, this route has been changed to `login/page.tsx` in order to more closely mimic a traditional SSR web server.

with regards to not requiring state anymore - while that is true, I think it would still be useful to illustrate how state works as some applications may like to use it (e.g. for returning users back to where they came from)

examples/nextjs-ssr/src/app/success/page.tsx
raynerljm: good catch! thanks :D

examples/nextjs-ssr/src/.DS_Store
raynerljm: 👍

raynerljm:
@kwajiehao with regards to your 1st point - yes! when you reviewed it the example had yet to be updated with PKCE (that's why it's WIP) - but now I've updated it with PKCE

with regards to the 2nd point - the error messages are not showing up because we are using `next start` to run the server and thus it's treated as a prod environment as the error messages are being omitted. I've searched around and I'm not sure that Next.js allows us to run `next start` while changing `NODE_ENV` to `development`. we could instead put change the message to prompt users to check their terminal for the error instead - what do you think?

.eslintrc
kwajiehao: Ah ok my bad, I didn't see any ESLint files being committed and I'm not super familiar with Next.js. I think we can leave it as is, I don't think this is super important, I was just curious haha

examples/nextjs-ssr/README.md
kwajiehao: Ahh ok, no issues then, just another instance of my lack of knowledge re: Next.js showing haha

examples/nextjs-ssr/src/app/api/auth-url/route.ts
kwajiehao: I think that makes sense! Changes in the README look good to me

examples/nextjs-ssr/src/app/api/auth-url/route.ts
kwajiehao: Right, sorry that wasn't clear. What I meant was that the value for the OAuth `state` is currently being generated by the "frontend" instead of the backend, but I see now that it's being used as a way of transmitting information (the ice cream flavor), which is fair. Let's keep it as-is then!

examples/nextjs-ssr/README.md
PrawiraGenestonlia: Next.js is making people dumb because they include everything ootb haha!

examples/nextjs-ssr/src/app/api/auth-url/route.ts
PrawiraGenestonlia: @kwajiehao I think it makes more sense to be param instead of routes because it is an extra optional parameters. Beside CSRF protection, anty wanted to showcase state as "save state - before and after AS" so yeah, even though we don't need it for PKCE, still using it to demonstrate save state. In this particular example, state refers to ice cream flavour.

PrawiraGenestonlia:
> with regards to the 2nd point - the error messages are not showing up because we are using `next start` to run the server and thus it's treated as a prod environment as the error messages are being omitted. I've searched around and I'm not sure that Next.js allows us to run `next start` while changing `NODE_ENV` to `development`. we could instead put change the message to prompt users to check their terminal for the error instead - what do you think?

This is next.js new features. It does not forward server error to FE through SSR unless we do special error boundary. Let's exclude this for this release because I don't think this is super important. Just some next.js shit features..

kwajiehao:
> with regards to the 2nd point - the error messages are not showing up because we are using `next start` to run the server and thus it's treated as a prod environment as the error messages are being omitted. I've searched around and I'm not sure that Next.js allows us to run `next start` while changing `NODE_ENV` to `development`. we could instead put change the message to prompt users to check their terminal for the error instead - what do you think?

Saw the changes you made - they look good to me! Don't think we need to do anything more beyond that. Thanks @raynerljm

kwajiehao (approved):
lgtm

PR #50, opened 2023-05-17
feat: express example

kwajiehao:
This is a rejig of #43 with a few major changes:
1. Frontend is no longer pulled and delivered statically, but run separately as a SPA (see https://github.com/opengovsg/sgid-demo-frontend-spa)
2. As part of that, configure CORS to work with the SPA
3. Update documentation
4. Update to use `sgid-client` v2.0.0, which uses PKCE by default


examples/express/README.md
mantariksh, line 1: do we need to add instructions here on how to run the frontend?

kwajiehao: We already have it in our docs (https://docs.id.gov.sg/integrations-with-sgid/typescript-javascript/express-with-single-page-app-frontend#step-5-test-it-out) but let me add it here too

kwajiehao: updated in [ca6c50a](https://github.com/opengovsg/sgid-client/pull/50/commits/ca6c50ab33b01daf5fe42ab8adf4121d81344a7c)

.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+examples
```
raynerljm: can i ask what is the difference between ignoring it in `.eslintignore` and using `ignorePatterns`?

examples/express/README.md
```diff
@@ -0,0 +1,62 @@
+# ExpressJS example sgID app
+
+The example application code is in `index.ts`. You can copy this code to bootstrap your sgID client application, and run this app locally to understand how this SDK helps you interact with the sgID server.
+
+## Running this example app locally
+
+### Prerequisites
+
+Register a new client at the [sgID developer portal](https://developer.id.gov.sg). For detailed instructions on how to register, check out our [documentation](https://jie-hao-kwa.gitbook.io/sgid-v2-beta/~/changes/75HA2ZSHqQMuHPNECzfs/introduction/getting-started/registration). Feel free to register a test client; there is no limit to the number of clients you can create.
```
raynerljm: we'll need to update the docs link haha

examples/express/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Steps to run locally
+
+1. Clone this repo.
+
+```
+git clone https://github.com/opengovsg/sgid-client.git
+```
+
+2. Go to this folder and copy the contents of `example.env` into a new file called `.env`.
```
raynerljm:
```suggestion
2. Go to this folder and copy the contents of `.env.example` into a new file called `.env`.
```

minor typo

examples/express/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+This should start the server on port `5001` by default
+
+5. Clone sgID's example frontend repo:
+
+```
+git clone https://github.com/opengovsg/sgid-demo-frontend-spa.git
+```
+
+6. Install and run the example frontend:
```
raynerljm, line 45: I just realized we need to add a step to ask them to copy the `.env.example` file into `.env` in the FE repo as well - I'll update that on GitBook

examples/express/index.ts
```diff
@@ -0,0 +1,170 @@
...
+  // Validate that the code verifier exists for this session
+  if (!session?.codeVerifier) {
+    res.redirect(`${frontendHost}/error`)
+    return
+  }
+
+  // Exchange the authorization code and code verifier for the access token
+  const { accessToken, sub } = await sgid.callback({
+    code: authCode,
+    nonce: session.nonce,
```
raynerljm: nit: should we destructure the `nonce` and `codeVerifier` from the session? because in`/userinfo` the values are being extracted from the session before being passed to `userinfo` (might also make passing the params a bit neater)

examples/express/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    accessToken,
+    sub,
+  })
+
+  // Add ice cream flavour to userinfo
+  userinfo.data.iceCream = session.state?.get('icecream') ?? 'None'
+  return res.json(userinfo)
+})
+
+apiRouter.get('/logout', async (_req, res) => {
```
raynerljm: should we clear delete the session as well when a user logs out?

examples/express/index.ts
```diff
@@ -0,0 +1,170 @@
...
+  try {
+    app.use(cookieParser())
+    app.use('/api', apiRouter)
+
+    app.listen(PORT, () => {
+      console.log(`Server listening on port ${PORT}`)
+      void open(`http://localhost:${PORT}`)
+    })
+  } catch (error) {
+    console.error(
```
raynerljm, line 166: we are no longer fetching the static files right? so I don't think we need this message anymore

raynerljm (commented):
mostly looks good! just some small nits - thanks for fixing this up! 🌮

.eslintignore
kwajiehao: There isn't a difference functionally according to their [docs](https://eslint.org/docs/latest/use/configure/ignore), so i'll remove this file. Thanks for pointing this out!

kwajiehao: Removed in [bf5470a](https://github.com/opengovsg/sgid-client/pull/50/commits/bf5470a124d74e0d4eca77d5b836be7c79b3b6c8)

examples/express/index.ts
kwajiehao: fixed in [930c29](https://github.com/opengovsg/sgid-client/pull/50/commits/930c29082d92ec6cbd0d6d48044c24ecce1ac3e7)

examples/express/README.md
kwajiehao: fixed in [930c29](https://github.com/opengovsg/sgid-client/pull/50/commits/930c29082d92ec6cbd0d6d48044c24ecce1ac3e7)

examples/express/README.md
kwajiehao: fixed in [930c29](https://github.com/opengovsg/sgid-client/pull/50/commits/930c29082d92ec6cbd0d6d48044c24ecce1ac3e7)

examples/express/index.ts
kwajiehao: fixed it in the callback code in [aca2c32](https://github.com/opengovsg/sgid-client/pull/50/commits/aca2c324b1dd3b1755a63a3a5c8568387a8754a1)

i didn't fix it for userinfo because you can't destructure an undefined item (`session` is possibly undefined). We could check for whether `session` is defined first, but since we also need to check for access token and sub, it doesn't really make sense to destructure there

examples/express/index.ts
kwajiehao: fixed in [a0992a4](https://github.com/opengovsg/sgid-client/pull/50/commits/a0992a443f2919f25008eaa091dc00187699dbb3)

examples/express/index.ts
```diff
@@ -145,7 +145,10 @@ apiRouter.get('/userinfo', async (req, res) => {
   return res.json(userinfo)
 })
 
-apiRouter.get('/logout', async (_req, res) => {
+apiRouter.get('/logout', async (req, res) => {
+  const sessionId = String(req.cookies[SESSION_COOKIE_NAME])
+  // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
```
raynerljm: nit: if we have to disable this rule, should we consider either turning off the rule or using javascript `Map`s instead?

examples/express/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "start": "ts-node index.ts",
+    "dev": "tsnd --transpile-only index.ts"
+  },
+  "dependencies": {
+    "@opengovsg/sgid-client": "^2.0.0",
+    "cookie-parser": "^1.4.6",
+    "cors": "^2.8.5",
+    "dotenv": "^16.0.3",
+    "express": "^4.18.2",
+    "extract-zip": "^2.0.1",
```
raynerljm: can we uninstall the dependencies that we no longer need because we are no longer fetching the static files?

raynerljm (approved):
just 2 minor suggestions! other than that LGTM :D

examples/express/index.ts
kwajiehao: Since `Map` is not fully supported by all browsers, I chose to turn off the rule instead. fixed in [beef837](https://github.com/opengovsg/sgid-client/pull/50/commits/beef837481ad8aa7430287c2c203bfe6055b1e07)

examples/express/package.json
kwajiehao: uninstalled in [f17059f0](https://github.com/opengovsg/sgid-client/pull/50/commits/f17059f065ebba7c120509e109679f7e29533322)

PR #52, opened 2023-05-23
fix(pkce): replace openid pkce generators with own implementations

### LINKED ISSUE #53: `generatePkcePair()` cannot run in browser environment
**Describe the bug**
openid-client is a Nodejs library that is not mean to run in browser environments. Since it is technically possible that RP's generate their PKCE pairs on the frontend, our implementation of PKCE pair generation should not depend on openid-client.

**To Reproduce**
1. `npm i @opengovsg/sgid-client` in a React project.
2. `import { generatePkcePair() } from @opengovsg/sgid-client` in a `.jsx`/`.tsx` file.
3. Run `const { codeVerifier, codeChallenge } = generatePkcePair()`
4. App crashes, with the following error message in console: 
image

**Expected behavior**
`generatePkcePair()` should run successfully and return `<codeChallenge, codeVerifier>` pair.

kschiew:
## Problem

`openid-client` is a Nodejs library that is not mean to run in browser environments. Since it is technically possible that RP's generate their PKCE pairs on the frontend, our implementation of PKCE pair generation should not depend on `openid-client`.
Closes #53 .

## Solution

Referenced (read: copied) `openid-client`'s implementation of PKCE pair generation

## Tests

WIP

### DIFF
--- a/src/browser/browserGenerators.ts
+++ b/src/browser/browserGenerators.ts
@@ -1,0 +1,71 @@
+import * as Errors from '../error'
+
+/**
+ * Generates a PKCE challenge pair where `codeChallenge` is the generated S256 challenge from `codeVerifier`
+ * @param length The length of the code verifier
+ * @returns The generated challenge pair
+ */
+export async function generatePkcePairBrowser(length = 43): Promise<{
+  codeVerifier: string
+  codeChallenge: string
+}> {
+  if (length < 43 || length > 128) {
+    throw new Error(Errors.PKCE_PAIR_LENGTH_ERROR)
+  }
+
+  const codeVerifier = generateCodeVerifierBrowser(length)
+  const codeChallenge = await generateCodeChallengeBrowser(codeVerifier)
+
+  return { codeVerifier, codeChallenge }
+}
+
+/**
+ * Generates the code verifier (random bytes encoded in url safe base 64) to be used in the OAuth 2.0 PKCE flow
+ * @param length The length of the code verifier to generate (Defaults to 43 if not provided)
+ * @returns The generated code verifier
+ */
+export function generateCodeVerifierBrowser(length = 43): string {
+  if (length < 43 || length > 128) {
+    throw new Error(Errors.CODE_VERIFIER_LENGTH_ERROR)
+  }
+
+  // 96 bytes results in a 128 long base64 string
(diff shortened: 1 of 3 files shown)


PrawiraGenestonlia (approved):
lgtm! will let @mantariksh do the final check.

mantariksh (commented):
hmm...it seems like we are committing to support `generatePkcePair` in browser environments. but if we are making this commitment, we should have corresponding unit tests.

currently our unit tests only use a Node environment (see `jest.config.js`). I think we should set up separate tests to test for the browser environment. there are a few ways to do this, but we can start here: https://jestjs.io/docs/configuration#testenvironment-string

wdyt?

PrawiraGenestonlia (changes requested):
This did not pass browser test for me.

PrawiraGenestonlia:
> hmm...it seems like we are committing to support `generatePkcePair` in browser environments. but if we are making this commitment, we should have corresponding unit tests.
> 
> 
> 
> currently our unit tests only use a Node environment (see `jest.config.js`). I think we should set up separate tests to test for the browser environment. there are a few ways to do this, but we can start here: https://jestjs.io/docs/configuration#testenvironment-string
> 
> 
> 
> wdyt?

Agree. I have a follow up question. Should we go with
1. Conditional import (import xx from "@opengovsg/sgid-client/browser)
2. Import path remains the same as node js but within the client, we will need to export accordingly - see jose package.

mantariksh:
another approach is to keep this repo for NodeJS, and have a completely separate one for browsers. advantage is separation of concerns and makes the package footprint smaller for client frontends.

raynerljm:
i agree with @mantariksh - i'm more inclined towards having a separate repo intended for browsers so as not to blur the lines between Node.js and browser environments

kwajiehao:
Discussed with @kschiew offline - my take is that we shouldn't make any rash decisions on updating our offering before we figure out whether we want to support a frontend PKCE generation flow. So in the meantime, we can use a third-party library like [pkce-challenge](https://www.npmjs.com/package/pkce-challenge) for the dev portal while Wira and Kok Seng figure out what the best practice here is

kschiew:
Closing PR until a new team decision is made. Pushed the code for custom implementation of browser-use PKCE generators for future reference.

PR #56, opened 2023-05-24
refactor: refactor examples to use redirect instead of callback

raynerljm:
## Problem

Non-standard naming (we use `callback` and `redirect` interchangeably in our dev docs/README/SDKs, etc.) In an effort to standardize the term, this PR changes all the examples to use `redirect` instead of `callback`.

This is done together with standardizing the [dev docs](https://docs.id.gov.sg/) and the [developer portal](https://github.com/datagovsg/sgid-developer-portal)

_How did you solve the problem?_

For Express example: 
- Change redirectUri definition in `SgidClient` to be `/redirect` instead of `/callback`
- Change the endpoint to `/redirect` from `/callback`

For Next.js CSR example:
- Change redirectUri definition in `SgidClient` to be `/redirect` instead of `/callback`
- Change the API route handler file name to `redirect.ts` from `callback.ts`


kwajiehao (approved):
lgtm

PR #57, opened 2023-05-24
docs: update example READMEs to point to the docs

raynerljm:
Updated the example READMEs to point to the docs instead of listing the steps.

Additionally renamed a test constant from `/callback`to `/redirect`as part of the effort to standardize the usage of our terms.

### DIFF
--- a/examples/express/README.md
+++ b/examples/express/README.md
@@ -4,57 +4,13 @@ The example application code is in `index.ts`. You can copy this code to bootstr
 
 ## Running this example app locally
 
-### Prerequisites
-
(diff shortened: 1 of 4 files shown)


mantariksh (commented):
wdyt about just pointing all the READMEs to the docs? same rationale as the main README, maintain a single source of truth

raynerljm:
pointed all the README to the docs following the Python [Flask example](https://github.com/opengovsg/sgid-client-python/tree/develop/examples/flask). additionally added an installation step in the 'Local development' section.

kwajiehao (approved):
lgtm

PR #58, opened 2023-05-25
chore: update port used for Next.js examples

kwajiehao:
This PR updates the port used for our Next.js CSR and SSR examples from `3000` to `5001`. 

`3000` is a commonly used port. We observed in our usability tests that when a user already has a process running on port `3000`, the Next.js example will then be run on a different port, and they will need to reconfigure quite a few things for the example to work (for example, they need to respecify their redirect URLs on both the dev portal and the sgID client initialization), which they might not know to do.

### DIFF
--- a/examples/nextjs-csr/README.md
+++ b/examples/nextjs-csr/README.md
@@ -17,4 +17,4 @@ npm install
 npm run dev
 ```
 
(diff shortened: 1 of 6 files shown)


mantariksh (approved):

raynerljm (approved):
lgtm! 🎉

PR #62, opened 2023-07-28
feat: expose generateNonce util

### LINKED ISSUE #61: Nonce API is inflexible and inaccurately typed
tl;dr there are 2 problems:
1. No util function to generate nonce
2. Nonce returned from `authorizationUrl` is inconveniently typed as `string | undefined`

I want to implement the following flow for generating the authorisation URL:
1. Controller calls a service to generate required params for authorisation URL (code verifier/challenge, nonce)
3. Controller calls another service to generate authorisation URL with the params generated in step 1
4. Controller saves code verifier and nonce in a cookie

But in step 1, there is no API to generate a nonce separately from auth URL. Workarounds are either:
1. Call `.authorizationUrl` just to get a random nonce. But this returns a nonce of type `string | undefined`, which forces me to write unnecessary null checks like `return nonce ?? ''`
2. Call `generateCodeVerifier` because it uses the same underlying function to generate the nonce. But this requires knowledge of sgID client internals.

Suggestion: expose a separate util function to generate a nonce, as with code verifier and code challenge.

mantariksh:
Closes #61 

## Alternatives considered

**Use string length as an argument instead of number of bytes**
My concern with this was that the user wants control over how random they want the nonce to be, and string length doesn't give them direct control over that. This is because the length can vary based on encoding rather than actual underlying randomness.

### DIFF
--- a/src/SgidClient.ts
+++ b/src/SgidClient.ts
@@ -1,5 +1,5 @@
 import { compactDecrypt, importJWK, importPKCS8 } from 'jose'
-import { Client, generators, Issuer } from 'openid-client'
+import { Client, Issuer } from 'openid-client'
 
 import {
   API_VERSION,
@@ -10,6 +10,7 @@ import {
   SGID_SUPPORTED_GRANT_TYPES,
 } from './constants'
 import * as Errors from './error'
+import { generateNonce } from './generators'
 import {
   AuthorizationUrlParams,
   AuthorizationUrlReturn,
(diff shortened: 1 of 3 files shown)


liangyuanruo:
FYI @kwajiehao @PrawiraGenestonlia the issue arose in [this](https://github.com/opengovsg/calsg/pull/467#discussion_r1277013922) PR comment, when @mantariksh was implementing SGID login for CalSG.

PrawiraGenestonlia (approved):
lgtm! Thanks @mantariksh !

PR #64, opened 2023-08-23
feat: handle array data values

kwajiehao:
## Problem
When we introduce POCDEX as a data source, it will return an array of public officer profiles associated with the NRIC / FIN provided. However, sgID cannot currently process array data values because of how data is stored on the mobile application - for Android the data is stored as a key-value map (so the value needs to be a string) and for iOS, the data is stored in binary form. Therefore, the data value for array fields need to be stored as JSON strings.

## Solution
This PR mitigates this problem by adding a new `parseData` method that parses the sgID user data values returned from the SgidClient's `userinfo` method into arrays or objects if applicable. It does so by detecting whether the data string is wrapped in square brackets. The method performs a safe JSON parse so that if the data is successfully parsed, an array object is returned, and if not, the original string is returned.

The reason for making the data parsing step an explicit, separate step is to avoid making breaking changes for our SDK. If we modify the return type of the `userinfo` method, this would break existing RP implementations, since existing users expect a string to be returned as a data value.

In practice, there's no breaking change because no array data fields are currently available in sgID. However, it would still represent a change in the types returned for the user. It is also a cleaner interface to consistently return a string and allow the user to parse it if necessary.

## Tests
- I created new unit tests for the util functions introduced in this PR
- I also ran the OIDC conformance test completely with no failed tests
- I ran the example app with the new client together with the `sgidthirdpartymock.public_officer_employments` scope

### DIFF
--- a/src/SgidClient.ts
+++ b/src/SgidClient.ts
@@ -16,7 +16,7 @@ import {
   AuthorizationUrlReturn,
   CallbackParams,
   CallbackReturn,
-  SafelyParsedJson,
+  ParsedSgidDataValue,
   SgidClientParams,
   UserInfoParams,
   UserInfoReturn,
@@ -212,7 +212,7 @@ export class SgidClient {
   private async decryptPayload(
     encryptedPayloadKey: string,
     data: Record<string, string>,
-  ): Promise<Record<string, SafelyParsedJson>> {
+  ): Promise<Record<string, string>> {
     let privateKeyJwk
     let payloadJwk
     try {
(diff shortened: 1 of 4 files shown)


src/SgidClient.ts
```diff
@@ -243,4 +248,19 @@ export class SgidClient {
...
+  /**
+   * Parses sgID user data.
+   * @param dataValue A value from the `data` object returned from the `userinfo` method
+   * @returns The parsed data value. If the input is a string, then a string is returned.
+   * If a stringified array or object is passed in, then an array or object is returned
+   * respectively.
+   */
+  parseData(dataValue: string): ParsedSgidDataValue {
+    // JSON parse array data values if necessary
+    if (isStringWrappedInSquareBrackets(dataValue)) {
```
PrawiraGenestonlia: Had an offline discussion with Jie Hao whether we should handle isStringWrappedInCurlyBrackets to parse object too.

PrawiraGenestonlia (approved):
left a comment for tracking. Other than that, lgtm!

src/SgidClient.ts
kwajiehao: As discussed, we should detect both stringified objects and stringified arrays. Modified in [ff2c39b0](https://github.com/opengovsg/sgid-client/pull/64/commits/ff2c39b01a79f7b156377256f594cc6845033850)

PrawiraGenestonlia (approved):
lgtm!

PR #68, opened 2023-09-11
feat: expose id token

kwajiehao:
## Overview

This PR does 2 things:
1. Expose the id token as part of the callback() method 
2. Perform validation on the id token in a manner consistent across SDKs

### 1. Expose the ID token as part of the `callback()` method

Currently, we only expose the `sub` and access token after performing all the requisite checks on the ID token. But the id token is useful because it acts as proof that the user has been authenticated by sgID. Some relying parties want to store
the ID token as part of their audit trail. This is why we're exposing the ID token here.

### 2. Perform validation on the id token in a manner consistent across SDKs

Validation is done according to the SDK implementation strategy here: https://www.notion.so/opengov/SDK-implementation-requirements-1f9b7cbd2bd4406b85d6645fe3e365dd 

Note that in this case, the `openid-client` library does validation on the ID token for us, so behavior deviates from the Python SDK.

## Tests

- Tested locally that the `idToken` is received with [sgid-demo-frontend-spa](https://github.com/opengovsg/sgid-demo-frontend-spa) and the Express example in this repo.
- Added unit tests

### DIFF
--- a/src/SgidClient.ts
+++ b/src/SgidClient.ts
@@ -160,7 +160,7 @@ export class SgidClient {
     const { sub } = tokenSet.claims()
     const { access_token: accessToken, id_token: idToken } = tokenSet
     if (!isNonEmptyString(idToken)) {
-      throw new Error(Errors.NO_ID_TOKEN_ERROR)
+      throw new Error(Errors.INVALID_ID_TOKEN_ERROR)
     }
     if (!sub) {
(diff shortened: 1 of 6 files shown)


PrawiraGenestonlia (approved):
lgtm!

PrawiraGenestonlia (approved):
lgtm!

PR #69, opened 2023-10-11
feat: parse sgid userinfo data

kwajiehao:
This PR adds a new method, `parseData`, which helps to parse the sgID userinfo object so that any stringified array or object values are parsed as objects. Previously, the `parseData` object only parsed individual values.

This PR is implemented according to the SDK implementation requirements as linked here: https://www.notion.so/opengov/SDK-implementation-requirements-1f9b7cbd2bd4406b85d6645fe3e365dd#dcabe67d14034525bfd5e84acc1c59c6
    
This PR also adds the necessary unit tests for the newly introduced utility functions that support `parseData`

### DIFF
--- a/src/SgidClient.ts
+++ b/src/SgidClient.ts
@@ -23,6 +23,7 @@ import {
 } from './types'
 import {
   convertPkcs1ToPkcs8,
+  isSgidUserinfoObject,
   isStringifiedArrayOrObject,
   safeJsonParse,
 } from './util'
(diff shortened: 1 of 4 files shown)


kwajiehao:
Current dependencies on/for this PR:
* develop
  * **PR #69** <a href="https://app.graphite.dev/github/pr/opengovsg/sgid-client/69" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opengovsg/sgid-client/69?utm_source=stack-comment).

src/util.ts
```diff
@@ -38,3 +38,32 @@ export function safeJsonParse(jsonString: string): ParsedSgidDataValue {
...
+}
+
+export function isSgidUserinfoObject(
+  data: unknown,
+): data is Record<string, string> {
+  if (!isDefinedObject(data)) {
+    return false
+  }
+
+  let result = true
```
PrawiraGenestonlia: nit: can consider this code instead of line 58-68 for code readability
```suggestion
  return Object.values(data).every((value) => typeof value === 'string')
```

src/SgidClient.ts
```diff
@@ -256,13 +257,34 @@ export class SgidClient {
...
+   * Parses the sgID user data object returned from the `userinfo` method.
+   * @param data A data object returned from the `userinfo` method
+   * @returns The parsed data object, where the value of each attribute is parsed as an
+   * object or array if applicable. If the value is a string, then a string is returned.
+   * If a stringified array or object is passed in, then an array or object is returned
+   * respectively.
+   */
+  parseData(data: unknown): Record<string, ParsedSgidDataValue> {
+    if (!isSgidUserinfoObject(data)) {
+      throw new Error(Errors.INVALID_SGID_USERINFO_DATA_ERROR)
```
PrawiraGenestonlia, line 269: For future consideration (nothing to do with this line): Not sure if we want to append sgID error into the original error traces before re-throwing because currently we throw custom error when catching openid-client error causing difficulties in debugging.

PrawiraGenestonlia (approved):
lgtm!

src/SgidClient.ts
kwajiehao: Agree, created [SGID-48](https://linear.app/ogp/issue/SGID-48/[sdk]-expose-full-error-trace-for-nodejs-sdk) on linear to track this

src/util.ts
kwajiehao: thanks, fixed in [b2060](https://github.com/opengovsg/sgid-client/pull/69/commits/b206067be2e7bf1841660a5561e75a757b2a711e)

btw for future suggested changes, you would need to drag to include the entire section that you want to replace - when i committed your suggestion directly, there was leftover code that needed to be removed

PR #70, opened 2023-10-16
Release v2.2.0

kwajiehao:
1. Expose ID token in callback function, as proof of authentication
2. Update the API for the `parseData` method


PrawiraGenestonlia (approved):
lgtm!

PR #87, opened 2025-01-07
v2.3.0

PrawiraGenestonlia:
Updating vulnerable packages


mantariksh (approved):
checked breaking changes in jose v5 as well, we shouldn't be affected: https://github.com/panva/jose/blob/main/CHANGELOG.md

